PR #10, opened 2023-05-26
Replace dynamic type checks with generic overloads

simonjbeaumont:
### Motivation

As described in https://github.com/apple/swift-openapi-generator/issues/41, there are a number of places where we using `as?` and `as!` type casting, which could be replaced with generic overloads.

### Modifications

- Replace converter helper functions with generic overloads.

### Result

- Resolves https://github.com/apple/swift-openapi-generator/issues/41.
- No visible impact.

### Test Plan

The tests continue to pass.

### Notes

Whilst implementing this change, I think that we may want to revisit the encoding logic, since I think the knowledge of whether to encode a value as binary data, JSON, or plain text, is at the place where the content-type is set, but we'll defer that to a subsequent PR.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/Converter+Client.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Converter+Client.swift
@@ -90,19 +90,27 @@ extension Converter {
         from data: Data,
         transforming transform: (T) -> C
     ) throws -> C {
-        let decoded: T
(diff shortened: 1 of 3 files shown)


simonjbeaumont:
I think that we'll close this for now, because it's not likely to be relevant if we tackle https://github.com/apple/swift-openapi-generator/issues/43.

PR #12, opened 2023-05-31
[Runtime] Choose the serialization method based on content type

czechboy0:
### Motivation

Resolves https://github.com/apple/swift-openapi-generator/issues/43. This PR also resolves https://github.com/apple/swift-openapi-generator/issues/42. It's paired with https://github.com/apple/swift-openapi-generator/pull/48.

This PR introduces the initial content-type-aware coding strategy, which is likely to get expanded in the future.

### Modifications

This PR covers the scope described in issue 43 - handling a `type: string` appropriately, based on whether it's wrapped in `application/json` or `text/plan` content.

It introduces new variants of the SPI helper functions, which now has variants per "coding strategy". It's best to read https://github.com/apple/swift-openapi-generator/pull/48/files?short_path=aa92e58#diff-aa92e58fb5a311512cfcd285827a4aa2e6634cae7fdfe0090bd2cfc7b0986140 for details.

Deprecated all of the SPI methods that didn't contain the coding strategy parameter, to be removed in the next breaking version.

### Result

We now have explicit control over the coding strategy for ambiguous types, such as `String`, which conform to both `Codable` and `LosslessStringConvertible`, to be taken advantage of by the generator (see the paired PR).

### Test Plan

Verified that the default behavior now matches version 0.1.0, and when paired with the updated generator, fully resolves issue 43.

Used code coverage to ensure that `Converter+{Common,Client,Server}.swift` are now 100% unit tested.


Sources/OpenAPIRuntime/Conversion/Converter+Client.swift
```diff
@@ -83,20 +83,25 @@ extension Converter {
...
-        if let myType = T.self as? _StringParameterConvertible.Type {
+        if let myType = T.self as? _StringParameterConvertible.Type,
+            strategy == .string
+        {
             guard
                 let stringValue = String(data: data, encoding: .utf8),
                 let decodedValue = myType.init(stringValue)
             else {
-                throw RuntimeError.failedToDecodePrimitiveBodyFromData
+                throw RuntimeError.failedToDecodeBody(type: T.self)
```
czechboy0, line 104: I liked this change of yours from #10, so ported it over, @simonjbeaumont.

Sources/OpenAPIRuntime/Conversion/Converter+Common.swift
```diff
@@ -55,7 +55,7 @@ public extension Array where Element == HeaderField {
     }
 }
 
-public extension Converter {
+extension Converter {
```
czechboy0, line 58: We were inconsistent in whether we put the access modifier on the extension or the methods, so I moved it to the methods. That also makes it easier to copy methods into `Deprecated.swift` when their time comes 😬

Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
```diff
@@ -281,17 +285,32 @@ public extension Converter {
...
+            guard
+                let stringValue = String(data: data, encoding: .utf8),
+                let decodedValue = myType.init(stringValue)
+            else {
+                throw RuntimeError.failedToDecodeBody(type: T.self)
+            }
+            decoded = decodedValue as! T
+        } else {
+            decoded = try decoder.decode(type, from: data)
+        }
```
simonjbeaumont, line 313: Should we consider changing the semantics here to be stricter?

Specifically, if `bodyGetRequired(...)` is called with `strategy: .string` then I think it should attempt to use that strategy and fail if that wasn't the right choice. 

Right now it looks like if we ask for `.string` but the type is not `_StringParameterConvertible` we fallback to using Codable.

Maybe what we want is something like

```swift
        switch strategy {
        case .string:
            guard let stringConvertibleType = T.self as? _StringParameterConvertible.Type else {
                preconditionFailure("...")
            }
            guard
                let stringValue = String(data: data, encoding: .utf8),
                let decodedValue = myType.init(stringValue)
            else {
                throw RuntimeError.failedToDecodeBody(type: T.self)
            }
            decoded = decodedValue as! T
        case .json:
            decoded = try decoder.decode(type, from: data)
        }
```

czechboy0: Possibly, although a few things to consider:
- the behavior had to be backwards compatible, so if you pass `.deferredToType`, you should get the "old" behavior
- the switch statement above doesn't handle `.deferredToType` - when you add it, you might see that you then have to duplicate the logic to decode both using string and codable methods inside that branch as well

simonjbeaumont: Hm, I think we can keep the backwards compatible functionality in the preserved functions that you've moved to `Deprecated.swift` and have this new function have the stricter semantics, especially for request and response body encoding because the content-type is explicitly mentioned in the OpenAPI document for each body.

czechboy0:
@simonjbeaumont Okay, based on our discussion, I split the `CodingStrategy` into two: `ParameterCodingStrategy` and `BodyCodingStrategy`, as they have different requirements. Ready for your re-review.

Sources/OpenAPIRuntime/Conversion/Converter+Client.swift
```diff
@@ -83,20 +83,25 @@ extension Converter {
...
         _ type: T.Type,
         from data: Data,
+        strategy: BodyCodingStrategy,
         transforming transform: (T) -> C
     ) throws -> C {
         let decoded: T
-        if let myType = T.self as? _StringParameterConvertible.Type {
+        if let myType = T.self as? _StringParameterConvertible.Type,
+            strategy == .string
+        {
```
simonjbeaumont, line 99: AFAICT this still has fallback semantics if called with `.string` when `T` does _not_ conform to `_StringParameterConvertible`, which IMO would be a bug in our generated code. As we discussed in the other thread, we think this method should be stricter since we have the more lenient method for backwards compatibility still.

Sources/OpenAPIRuntime/Conversion/Converter+Client.swift
```diff
@@ -139,9 +144,11 @@ extension Converter {
     ) throws -> Data {
         let body = transform(value)
         headerFields.add(name: "content-type", value: body.contentType)
-        if let value = value as? _StringParameterConvertible {
+        if let value = value as? _StringParameterConvertible,
+            body.strategy == .string
+        {
```
simonjbeaumont, line 149: Ditto about strictness.

Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
simonjbeaumont: Did we address this in the latest commit?

simonjbeaumont (changes requested):
Splitting out the enum is great, but I thought we were also going to make the new methods more strict and keep the fallback behaviour the backwards compatible functions you moved to Deprecated?

czechboy0:
@simonjbeaumont This is now ready. It's exactly what we discussed and already codified requirements for in https://github.com/apple/swift-openapi-generator/pull/50

Now, this is a large PR and is by no means perfect, but it's a solid move in the right direction. There are things I'd like to improve in the future, but I wouldn't want to hold this PR open for it.

It's very repetitive, as we're doing ~50 variants of similar logic, but I took special care, and all the methods are unit tested and used in our integration test in the generator repo, so I'm confident that this logic is at least as good as what we have in 0.1.0.

Sources/OpenAPIRuntime/Conversion/Converter+Client.swift
czechboy0: Redesigned in the latest approach.

simonjbeaumont (approved):
Great, thanks @czechboy0!

PR #13, opened 2023-06-06
Rename _StringParameterConvertible to _StringConvertible

czechboy0:
### Motivation

As part of reworking https://github.com/apple/swift-openapi-runtime/pull/12, we'll be using the protocol `_StringParameterConvertible` not just for parameters, but also for bodies, so renaming it accordingly to reflect that.

### Modifications

Renamed `_StringParameterConvertible` to `_StringConvertible`.

### Result

The naming is more accurate now.

Compatibility-wise, this isn't affecting the generator, as this protocol is not used directly by generated code, only indirectly through `_AutoLosslessStringConvertible`, so we don't even need to stage this change.

### Test Plan

All tests pass.


simonjbeaumont (approved):
LGTM. Marking as implementation-only, assuming that's correct?

czechboy0:
Yup, while this is a public symbol, it's an underscored marker protocol that nothing external should be using, so the same guarantees as SPI.

PR #19, opened 2023-06-14
Correct variable name in `ClientMiddleware` reference documentation

sikevux:
### Motivation

The example for `ClientMiddleware` doesn't compile

### Modifications

Correct the variable name in `ClientMiddleware`

### Result

The example can now compile

### DIFF
--- a/Sources/OpenAPIRuntime/Interface/ClientTransport.swift
+++ b/Sources/OpenAPIRuntime/Interface/ClientTransport.swift
@@ -207,7 +207,7 @@ public protocol ClientTransport: Sendable {
 ///         ) async throws -> Response {
 ///             var request = request
 ///             request.headerFields.append(.init(
-///                 name: "Authorization", value: "Bearer \(token)"
+///                 name: "Authorization", value: "Bearer \(self.bearerToken)"
 ///             ))
 ///             return try await next(request, baseURL)
 ///         }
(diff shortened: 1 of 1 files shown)


simonjbeaumont (approved):
Thanks @sikevux—nice catch!

Sources/OpenAPIRuntime/Interface/ClientTransport.swift
```diff
@@ -207,7 +207,7 @@ public protocol ClientTransport: Sendable {
 ///         ) async throws -> Response {
 ///             var request = request
 ///             request.headerFields.append(.init(
-///                 name: "Authorization", value: "Bearer \(token)"
+///                 name: "Authorization", value: "Bearer \(self.bearerToken)"
```
czechboy0, line 210:
```suggestion
///                 name: "Authorization", value: "Bearer \(bearerToken)"
```

We generally omit the `self.` unless required by the compiler.

czechboy0 (changes requested):
Thanks @sikevux, just one small suggestion.

Sources/OpenAPIRuntime/Interface/ClientTransport.swift
simonjbeaumont: Personally I think we _should_ use explicit `self.` because it helps with local reasoning and avoids unintended shadowing.

However, @czechboy0 is right that it's not currently a convention in this repo so, if we do move to it, we'll do so holistically, across the whole codebase, with the aid of `swift-format`.

So, with that in mind, I think we should take this suggestion.

czechboy0 (approved):
Thanks! 🙏

czechboy0:

PR #20, opened 2023-06-16
Converter SPI methods for required fields should return non-optional types

sikevux:
### Motivation

`getRequiredRequestBodyAsText` currently returns `C?` instead of `C`. As the value should be required an optional in the return is causing issues.

The errors are `Cannot convert value of type 'C?' to specified type` and `Cannot infer contextual base in reference to member 'text'`

 ### Modifications

Change the return type of `getRequiredRequestBodyAsText`

 ### Result

Code using a required requestBody now compiles without errors.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/Converter+Common.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Converter+Common.swift
@@ -163,7 +163,7 @@ extension Converter {
         in headerFields: [HeaderField],
         name: String,
         as type: [T].Type
-    ) throws -> [T]? {
+    ) throws -> [T] {
         try getRequiredHeaderFields(
             in: headerFields,
             name: name,
@@ -219,7 +219,7 @@ extension Converter {
         in headerFields: [HeaderField],
         name: String,
         as type: [Date].Type
-    ) throws -> [Date]? {
+    ) throws -> [Date] {
         try getRequiredHeaderFields(
             in: headerFields,
             name: name,
--- a/Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
@@ -202,7 +202,7 @@ public extension Converter {
         _ type: T.Type,
         from data: Data?,
         transforming transform: (T) -> C
-    ) throws -> C? {
+    ) throws -> C {
         try getRequiredRequestBody(
             type,
             from: data,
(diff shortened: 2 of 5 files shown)


czechboy0:
Thanks for the fix, this was an oversight on my part.

czechboy0 (approved):

simonjbeaumont:

simonjbeaumont:
@sikevux this is great, thank you. Is it possible to add a test case for this? Specifically something that uses the code that previously world not compile?

czechboy0:
Agreed, we have tests already, but adding one that does something like

`let foo: String = try get...` should do the trick - should not compile without the fix, and should with.

czechboy0 (changes requested):
Marking as needs work based on the conversation above.

czechboy0:
Ugh, I took a quick look and we have the same issue in a few more places:
- https://github.com/apple/swift-openapi-runtime/blob/bec8d38be0254a59baca6e6434823f77cf1d734c/Sources/OpenAPIRuntime/Conversion/Converter%2BCommon.swift#L166
- https://github.com/apple/swift-openapi-runtime/blob/bec8d38be0254a59baca6e6434823f77cf1d734c/Sources/OpenAPIRuntime/Conversion/Converter%2BCommon.swift#L222
- https://github.com/apple/swift-openapi-runtime/blob/bec8d38be0254a59baca6e6434823f77cf1d734c/Sources/OpenAPIRuntime/Conversion/Converter%2BServer.swift#L205

Can you please fix those too? 🙏

sikevux:
> Ugh, I took a quick look and we have the same issue in a few more places:
> 
> * https://github.com/apple/swift-openapi-runtime/blob/bec8d38be0254a59baca6e6434823f77cf1d734c/Sources/OpenAPIRuntime/Conversion/Converter%2BCommon.swift#L166
> * https://github.com/apple/swift-openapi-runtime/blob/bec8d38be0254a59baca6e6434823f77cf1d734c/Sources/OpenAPIRuntime/Conversion/Converter%2BCommon.swift#L222
> * https://github.com/apple/swift-openapi-runtime/blob/bec8d38be0254a59baca6e6434823f77cf1d734c/Sources/OpenAPIRuntime/Conversion/Converter%2BServer.swift#L205
> 
> Can you please fix those too? 🙏

Yeah. Modified all the `get` tests to explicitly set the type just to make sure there's no more. Seems to be all of them now 😸

czechboy0 (approved):
Great, thank you!

simonjbeaumont (approved):
Awesome. Thanks @sikevux!

PR #22, opened 2023-07-06
Better Sendability to avoid warnings when -strict-concurrency = targeted+

MahdiBM:
### Motivation

When using `-strict-concurrency=targeted`, `ClientMiddleware` will emit warnings if an actor is conformed to it:
```swift
import OpenAPIRuntime
import Foundation

actor Middleware: ClientMiddleware {
    func intercept(
        _ request: Request,
        baseURL: URL,
        operationID: String,
        /// Need to add `@Sendable` or the warnings don't go away even with this PR. Code-completion won't add it.
        next: @Sendable (Request, URL) async throws -> Response
    ) async throws -> Response {
        try await next(request, baseURL)
    }
}
```
The code above emit the following warning with the current implementation:

> Sendability of function types in instance method 'intercept(_:baseURL:operationID:next:)' does not match requirement in protocol 'ClientMiddleware'

Repro package: https://github.com/MahdiBM/OAPIRClientWarningsRepro

You can change between this PR and the main branch by commenting out the dependency declarations i've already put in the `Package.swift`.
Then observe there are no warnings with this PR, unlike with the main branch.

### Modifications

Generally add more @Sendables everywhere. Also use @preconcurrency on Sendable functions to ease the migration:
Currently we have:
```swift
public protocol ClientMiddleware: Sendable {
    func intercept(
        _ request: Request,
        baseURL: URL,
        operationID: String,
        next: (Request, URL) async throws -> Response
    ) async throws -> Response
}
```
With this PR we'll have:
```swift
public protocol ClientMiddleware: Sendable {
    @preconcurrency
    /// ~~~^ notice `@preconcurrency`
    func intercept(
        _ request: Request,
        baseURL: URL,
        operationID: String,
        next: @Sendable (Request, URL) async throws -> Response
        /// ~~~^ notice `@Sendable`
    ) async throws -> Response
}
```

### Result

No more warnings when using `-strict-concurrency=targeted` then conforming an actor to `ClientMiddleware`.

### Test Plan

Manually tested, and there is a [repro package](https://github.com/MahdiBM/OAPIRClientWarningsRepro) for maintainers to check.

### Considerations

I've gone an step ahead and added `@preconcurrency` too, to the functions with `@Sendable` closure.
I did it for the compiler to emit warnings instead of throwing errors when users pass a non-sendable closure value as the supposedly-sendable closure.
I can remove those if necessary.

### DIFF
--- a/Sources/OpenAPIRuntime/Interface/ClientTransport.swift
+++ b/Sources/OpenAPIRuntime/Interface/ClientTransport.swift
@@ -229,10 +229,11 @@ public protocol ClientMiddleware: Sendable {
     ///   - operationID: The identifier of the OpenAPI operation.
     ///   - next: A closure that calls the next middleware, or the transport.
     /// - Returns: An HTTP response.
+    @preconcurrency
     func intercept(
         _ request: Request,
         baseURL: URL,
         operationID: String,
-        next: (Request, URL) async throws -> Response
+        next: @Sendable (Request, URL) async throws -> Response
     ) async throws -> Response
 }
--- a/Sources/OpenAPIRuntime/Interface/ServerTransport.swift
+++ b/Sources/OpenAPIRuntime/Interface/ServerTransport.swift
@@ -111,10 +111,9 @@ public protocol ServerTransport {
     ///   - path: The URL path components, for example `["pets", ":petId"]`.
     ///   - queryItemNames: The names of query items to be extracted
     ///   from the request URL that matches the provided HTTP operation.
+    @preconcurrency
     func register(
-        _ handler: @Sendable @escaping (
-            Request, ServerRequestMetadata
-        ) async throws -> Response,
+        _ handler: @Sendable @escaping (Request, ServerRequestMetadata) async throws -> Response,
         method: HTTPMethod,
         path: [RouterPathComponent],
         queryItemNames: Set<String>
@@ -215,10 +214,11 @@ public protocol ServerMiddleware: Sendable {
     ///   - operationID: The identifier of the OpenAPI operation.
     ///   - next: A closure that calls the next middleware, or the transport.
     /// - Returns: An HTTP response.
+    @preconcurrency
     func intercept(
         _ request: Request,
         metadata: ServerRequestMetadata,
         operationID: String,
-        next: (Request, ServerRequestMetadata) async throws -> Response
+        next: @Sendable (Request, ServerRequestMetadata) async throws -> Response
     ) async throws -> Response
 }
--- a/Sources/OpenAPIRuntime/Interface/UniversalClient.swift
+++ b/Sources/OpenAPIRuntime/Interface/UniversalClient.swift
@@ -83,12 +83,14 @@ public struct UniversalClient: Sendable {
     ///   - serializer: Creates an HTTP request from the provided Input value.
     ///   - deserializer: Creates an Output value from the provided HTTP response.
     /// - Returns: The Output value produced by `deserializer`.
+    @preconcurrency
     public func send<OperationInput, OperationOutput>(
         input: OperationInput,
         forOperation operationID: String,
-        serializer: (OperationInput) throws -> Request,
-        deserializer: (Response) throws -> OperationOutput
+        serializer: @Sendable (OperationInput) throws -> Request,
+        deserializer: @Sendable (Response) throws -> OperationOutput
     ) async throws -> OperationOutput {
+        @Sendable
         func wrappingErrors<R>(
             work: () async throws -> R,
             mapError: (Error) -> Error
@@ -121,7 +123,7 @@ public struct UniversalClient: Sendable {
             makeError(error: error)
         }
         let response: Response = try await wrappingErrors {
-            var next: (Request, URL) async throws -> Response = { (_request, _url) in
+            var next: @Sendable (Request, URL) async throws -> Response = { (_request, _url) in
                 try await wrappingErrors {
                     try await transport.send(
                         _request,
--- a/Sources/OpenAPIRuntime/Interface/UniversalServer.swift
+++ b/Sources/OpenAPIRuntime/Interface/UniversalServer.swift
@@ -84,14 +84,16 @@ public struct UniversalServer<APIHandler: Sendable>: Sendable {
     ///   - deserializer: Creates an Input value from the provided HTTP request.
     ///   - serializer: Creates an HTTP response from the provided Output value.
     /// - Returns: The HTTP response produced by `serializer`.
+    @preconcurrency
     public func handle<OperationInput, OperationOutput>(
         request: Request,
         with metadata: ServerRequestMetadata,
         forOperation operationID: String,
-        using handlerMethod: @escaping (APIHandler) -> ((OperationInput) async throws -> OperationOutput),
-        deserializer: @escaping (Request, ServerRequestMetadata) throws -> OperationInput,
-        serializer: @escaping (OperationOutput, Request) throws -> Response
+        using handlerMethod: @Sendable @escaping (APIHandler) -> ((OperationInput) async throws -> OperationOutput),
+        deserializer: @Sendable @escaping (Request, ServerRequestMetadata) throws -> OperationInput,
+        serializer: @Sendable @escaping (OperationOutput, Request) throws -> Response
     ) async throws -> Response {
+        @Sendable
         func wrappingErrors<R>(
             work: () async throws -> R,
             mapError: (Error) -> Error
@@ -102,6 +104,7 @@ public struct UniversalServer<APIHandler: Sendable>: Sendable {
                 throw mapError(error)
             }
         }
+        @Sendable
         func makeError(
             input: OperationInput? = nil,
             output: OperationOutput? = nil,
@@ -116,7 +119,7 @@ public struct UniversalServer<APIHandler: Sendable>: Sendable {
                 underlyingError: error
             )
         }
-        var next: (Request, ServerRequestMetadata) async throws -> Response = { _request, _metadata in
+        var next: @Sendable (Request, ServerRequestMetadata) async throws -> Response = { _request, _metadata in
             let input: OperationInput = try await wrappingErrors {
                 try deserializer(_request, _metadata)
             } mapError: { error in


czechboy0:
@MahdiBM Is this PR ready for us to review?

czechboy0:

MahdiBM:
@czechboy0 yes 🙂

czechboy0:
@MahdiBM Is this agains an option we could enable in Package.swift for the whole library, to prepare us for Swift 6, similar to https://github.com/apple/swift-openapi-generator/pull/99? If we're making these changes, I'd like to make sure we don't regress it in the future.

MahdiBM:
So adding @Sendable can only help for better concurrency checks.

Adding @preconcurrency though, we won't be able to remove it without it counting as a Breaking Change, which means it needs a major release, and that'll probably be when Swift 6 comes out, because Swift 6 which comes with the default settings of basically `-strict-concurrency=complete`.

@preconcurrency is useful to ease people's migration to use concurrency stuff. It basically turns Sendable errors of declarations to just warnings.

I personally don't need it as my projects are for the most part already in compliance with at least `-strict-concurrency=targeted`. I reckon it might not be needed. I'm not sure at all. I'm curious to know what @FranzBusch thinks.

FranzBusch:
> Adding @preconcurrency though, we won't be able to remove it without it counting as a Breaking Change

That is not true. You can free add and remove that attribute whenever required. Especially if you version guard it. Overall, I agree though we should try to be strict concurrency warning free in this library. 

@czechboy0 We are currently in the process of turning on strict concurrency checking in our CI in all our NIO libraries so feel free to add it here on the CI as well. You can simple pass this flag `-strict-concurrency=complete` to your swift invocation.

Sources/OpenAPIRuntime/Interface/ClientTransport.swift
```diff
@@ -229,10 +229,11 @@ public protocol ClientMiddleware: Sendable {
     ///   - operationID: The identifier of the OpenAPI operation.
     ///   - next: A closure that calls the next middleware, or the transport.
     /// - Returns: An HTTP response.
+    @preconcurrency
```
FranzBusch, line 232: We should not add this and any of the other `@preconcurrency` attributes. Just marking the closure as `@Sendable` seems all we need.

czechboy0:
If we can go all the way to `complete` in one go, I'd prefer that as well. We'll have to see if there are any non-trivial things to fix up in the runtime library, and add the flag to the docker-compose file so that CI picks it up. @MahdiBM, can you try to see if `complete` works as well? Then we can get this PR into a shape where it passes all CI without warnings.

MahdiBM:
@FranzBusch 
> You can free add and remove that attribute whenever required

is it not true that removing `@preconcurrency` from a function will make it so some warnings turn into errors? How is that acceptable without a major release?

FranzBusch:
> is it not true that removing @preconcurrency from a function will make it so some possible warnings turn into errors? How is that acceptable without a major release?

AFAIK, this will only add warnings and not errors. So it is safe to remove this. However, I don't think we should or need these attributes here in the first place

MahdiBM:
Yeah, for this PR i'll be removing the @preconcurrencys. I'm just trying to see what i'm wrong about.

Imagine having this. It'll show a warning:
```swift
struct NotSendable {
    var a = ""
}

@preconcurrency
public func something(_: @Sendable () -> Void) { }

func another() async throws {
    var notSendable = NotSendable()
    something {
        notSendable.a = ""
///     ^~~~ Warning: Mutation of captured var 'notSendable' in concurrently-executing code; this is an error in Swift 6
    }
}
```
If you remove the `@preconcurrency`, the warning will turn into an error:
```swift
public func something(_: @Sendable () -> Void) { }

func another() async throws {
    var notSendable = NotSendable()
    something {
        notSendable.a = ""
///     ^~~~ Error: Mutation of captured var 'notSendable' in concurrently-executing code
    }
}
```
So if we publish the package with functions having @preconcurrency, then someone writes some code that has warning like above, removing @preconcurrency will make it so those warnings turn into errors ... I think that qualifies as a "breaking change"?

MahdiBM:
@czechboy0 
> If we can go all the way to complete in one go, I'd prefer that as well. We'll have to see if there are any non-trivial things to fix up in the runtime library, and add the flag to the docker-compose file so that CI picks it up. @MahdiBM, can you try to see if complete works as well? Then we can get this PR into a shape where it passes all CI without warnings.

One thing to note is that we'll  _need_ to add the flag only in CI files like you requested. We can add it as `.unsafeFlags(["-strict-concurrency=targeted"])` in Swift-settings of targets, but that'll throw weird errors for other libraries that use this package, if they rely on any tagged release. The good news is a fix for that is already merged and should be coming with Swift 5.9, and at that point we can just use that in the Package.swift: https://github.com/apple/swift/pull/66991

That being said, I'll check to see if we can use that flag without too many changes.

MahdiBM:
@czechboy0 This should be ready.

- The `RedactedHeadersStorage`'s purpose is to only silence strict-concurrency warnings that are due to mutation of static properties.

@FranzBusch
 
> the reason why this is not breaking is that adopters of your package can always apply @preconcurrency on your modules 

Honestly that might have been smart to have, but i just checked, and you can't silence the errors from functions of another module, using @preconcurrency on the the import.

> I don't think it is intended that this silences the error you are seeing without @preconcurrency

I think it's intended as we figured in the ongoing work of @0xTim to make Vapor stuff more Sendable.
It helps converting the old functions that have not-@Sendable closures, to functions with @Sendable closures. It enabled us to make those closures @Sendable, while marking the functions as `@preconcurrency` so the newly-`@Sendable`-marked closure doesn't result in code-breakage of users (as we learnt the hard way when users complained about their code breakages)

FranzBusch:
> I think it's intended as we figured in the ongoing work of @0xTim to make Vapor stuff more Sendable.
It helps converting the old functions that have not-https://github.com/sendable closures, to functions with https://github.com/sendable closures. It enabled Tim to make those closures https://github.com/sendable, while marking the functions as @preconcurrency so the newly-@Sendable-marked closures don't result in code-breakage of users (as we learnt the hard way when users complained about their code breakages)

Sure for existing code this might be the case and you have to handle this. However, this package has not yet hit a 1.0.0 and we can safely introduce this without adding `@preconcurrency`

MahdiBM:
> However, this package has not yet hit a 1.0.0 and we can safely introduce this without adding `@preconcurrency`
Yeah you're right in that sense, i've already removed all the @preconcurrencys 🙂

czechboy0:
@MahdiBM, please also add the CLI flag for the strict concurrency checking to the CI script (the docker-compose file), to see that it's all passing before we start reviewing the PR again, thanks! 🙏

Sources/OpenAPIRuntime/Interface/CurrencyTypes.swift
```diff
@@ -47,20 +82,12 @@ extension HeaderField {
     /// Use this to avoid leaking sensitive tokens into application logs.
     public static var redactedHeaderFields: Set<String> {
```
FranzBusch, line 83: This is actually a bit surprising to me. This is a global static which can be set to redact header fields? This sounds to be like something that should be rather passed in a configuration struct. @czechboy0 @simonjbeaumont is this something you considered before?

FranzBusch: Got you. We had a similar discussion recently in AHC https://github.com/swift-server/async-http-client/issues/689.

I don't know if I share the safety concern of including everything in `CustomStringConvertible`. If somebody `print()` a request or response then they must expect that there are sensitive informations in there. This can not only be in the headers but also in the path.

Personally, I would lean towards including everything in the `description` and it is up to the user if they log a request or not. If they do log a request they are themselves responsible for making sure it doesn't include sensitive data as long as we don't log it ourselves I don't see us in the responsibility to sanitise those types. Happy to move this discussion to a different place though.

simonjbeaumont: I also don't love that we do this and see this as a bit of a layering violation and the filtering could easily be achieved in a logging middleware. 

My biggest concern is it implies "safe default", which is really best-effort at best.

czechboy0: I appreciate these thoughts, but as @FranzBusch suggested, let's move this to a dedicated issue/thread. This PR is about passing stricter concurrency checking.

MahdiBM:
@czechboy0 added the flag but i figure it should probably be an argument like `$${STRICT_CONCURRENCY_ARG-}` if you want to keep it for merge too. let me know if i should just delete it for now after a round of tests, or you want to use the $$ form.

MahdiBM:
There seem to be some problems caused by the fact that Sendable conformances of Foundation types on Linux and macOS are not the same. I'll try to resolve those for now. I'm guessing we need to wait for swift-foundation until these differences are resolved by Foundation itself.

MahdiBM:
Seems ready now 🙂

Sources/OpenAPIRuntime/Interface/CurrencyTypes.swift
simonjbeaumont: Sure—not suggesting we derail this PR. I was mentioned in this thread for an opinion and there is no other issue currently, so felt rude not to respond when asked 😕

Sources/OpenAPIRuntime/Interface/CurrencyTypes.swift
```diff
@@ -17,6 +17,40 @@ import Foundation
...
+    }
+
+    /// The lock used for protecting access to `_locked_redactedHeaderFields`.
+    private let lock: NSLock = {
+        let lock = NSLock()
+        lock.name = "com.apple.swift-openapi-runtime.lock.redactedHeaderFields"
+        return lock
+    }()
+
+    init() {}
```
simonjbeaumont, line 51:
```suggestion
```

Isn't this implicit?

docker/docker-compose.yaml
```diff
@@ -30,7 +30,7 @@ services:
 
   test:
     <<: *common
-    command: /bin/bash -xcl "swift $${SWIFT_TEST_VERB-test} $${WARN_AS_ERROR_ARG-} $${SANITIZER_ARG-} $${IMPORT_CHECK_ARG-}"
+    command: /bin/bash -xcl "swift $${SWIFT_TEST_VERB-test} $${WARN_AS_ERROR_ARG-} $${SANITIZER_ARG-} $${IMPORT_CHECK_ARG-} -Xswiftc -strict-concurrency=complete"
```
simonjbeaumont, line 33: To follow the convention here we should probably replace this with...
```suggestion
    command: /bin/bash -xcl "swift $${SWIFT_TEST_VERB-test} $${WARN_AS_ERROR_ARG-} $${SANITIZER_ARG-} $${IMPORT_CHECK_ARG-} $${STRICT_CONCURRENCY_ARG-}"
```

...and then specify it in:
- `docker/docker-compose.2204.58.yaml`
- `docker/docker-compose.2204.59.yaml`

If you look in those files, that's how `-Xswiftc -warnings-as-errors` is passed.

simonjbeaumont (changes requested):
Thanks for doing this. Functionally this LGTM. Just a few nits.

Sources/OpenAPIRuntime/Interface/CurrencyTypes.swift
czechboy0: Yeah not suggesting this was derailing it, just that this is a good discussion and we should preserve the points in a separate issue, ideally with someone suggesting what we should change it _to_, with arguments of how that's overall a better solution. 🙂

simonjbeaumont: Moved to https://github.com/apple/swift-openapi-generator/issues/113.

docker/docker-compose.yaml
MahdiBM: I did it for `docker-compose.2204.main.yaml` too as i couldn't think of a reason why that file should be left out. Let me know if it shouldn't be there.

simonjbeaumont (approved):
Thanks @MahdiBM

simonjbeaumont:
@MahdiBM I updated the pull request title. Would you mind updating the description to reflect the changes made during review, as it will be used in the merge commit. 🙏

MahdiBM:
@simonjbeaumont does this seem good?

Sources/OpenAPIRuntime/Errors/ClientError.swift
```diff
@@ -11,7 +11,11 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 //===----------------------------------------------------------------------===//
+#if canImport(Darwin)
 import Foundation
+#else
+@preconcurrency import Foundation
```
FranzBusch, line 17: For all the places where we do this could we specify what type we are importing here by doing `import struct Foundation.URL`. This way we know what we need to make `Sendable` and it narrows down the preconcurrency import a bit more

simonjbeaumont: This is just copied from the other places we import Foundation in this package.

I think this is a great piece of feedback but I don't want to abuse @MahdiBM's good will. I can create an issue for this and take a look.

MahdiBM: I already did it but i haven't actually confirmed/tested the comments.
I added comments instead of e.g. `import struct Foundation.URL` because the `URL` wasn't the only thing needed from `Foundation`.

simonjbeaumont: I filed https://github.com/apple/swift-openapi-generator/issues/114 if we need any followup.

Sources/OpenAPIRuntime/Base/OpenAPIValue.swift
```diff
@@ -327,7 +327,7 @@ public struct OpenAPIObjectContainer: Codable, Equatable, Hashable, Sendable {
 
     public func encode(to encoder: Encoder) throws {
         var container = encoder.singleValueContainer()
-        try container.encode(value.mapValues(OpenAPIValueContainer.init))
+        try container.encode(value.mapValues(OpenAPIValueContainer.init(unvalidatedValue:)))
```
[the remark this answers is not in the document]
MahdiBM: No honestly. I just thought you probably weren't using the validated init considering the data could be any json coming from outside.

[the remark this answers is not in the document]
MahdiBM: Ah this is the encode function. Then I'm wrong. It should probably be the other init.

czechboy0:
@MahdiBM please lmk once the CI is green again and I'll do one more review 🙏

MahdiBM:
@czechboy0 i think this change is just a breaking change so i can't really make the CI green 😅:
```bash
15 breaking changes detected in OpenAPIRuntime:
  💔 API breakage: constructor OpenAPIValueContainer.init(unvalidatedValue:) has parameter 0 type change from Any? to (Swift.Sendable)?
  💔 API breakage: var ClientError.operationInput has declared type change from Any to Swift.Sendable
  💔 API breakage: accessor ClientError.operationInput.Get() has return type change from Any to Swift.Sendable
  💔 API breakage: accessor ClientError.operationInput.Set() has parameter 0 type change from Any to Swift.Sendable
  💔 API breakage: constructor ClientError.init(operationID:operationInput:request:baseURL:response:underlyingError:) has parameter 1 type change from Any to Swift.Sendable
  💔 API breakage: var ServerError.operationInput has declared type change from Any? to (Swift.Sendable)?
  💔 API breakage: accessor ServerError.operationInput.Get() has return type change from Any? to (Swift.Sendable)?
  💔 API breakage: accessor ServerError.operationInput.Set() has parameter 0 type change from Any? to (Swift.Sendable)?
  💔 API breakage: var ServerError.operationOutput has declared type change from Any? to (Swift.Sendable)?
  💔 API breakage: accessor ServerError.operationOutput.Get() has return type change from Any? to (Swift.Sendable)?
  💔 API breakage: accessor ServerError.operationOutput.Set() has parameter 0 type change from Any? to (Swift.Sendable)?
  💔 API breakage: constructor ServerError.init(operationID:request:requestMetadata:operationInput:operationOutput:underlyingError:) has parameter 3 type change from Any? to (Swift.Sendable)?
  💔 API breakage: constructor ServerError.init(operationID:request:requestMetadata:operationInput:operationOutput:underlyingError:) has parameter 4 type change from Any? to (Swift.Sendable)?
  💔 API breakage: func UniversalClient.send(input:forOperation:serializer:deserializer:) has generic signature change from <OperationInput, OperationOutput> to <OperationInput, OperationOutput where OperationInput : Swift.Sendable, OperationOutput : Swift.Sendable>
  💔 API breakage: func UniversalServer.handle(request:with:forOperation:using:deserializer:serializer:) has generic signature change from <APIHandler, OperationInput, OperationOutput where APIHandler : Swift.Sendable> to <APIHandler, OperationInput, OperationOutput where APIHandler : Swift.Sendable, OperationInput : Swift.Sendable, OperationOutput : Swift.Sendable>
```
They are all complaining about Sendable.

czechboy0:
Ah ofc, I didn't realize. Let me talk with @simonjbeaumont tomorrow about how we'll land and release this breaking change.

simonjbeaumont:
> Ah ofc, I didn't realize. Let me talk with @simonjbeaumont tomorrow about how we'll land and release this breaking change.

I already labelled this PR as `semver/minor` so I think we just release and cut a new version. We can highlight in the release notes that this is breaking and we already advise in our docs that folks should be using `upToNextMinor` during the pre-1.0 period if they want to avoid breaking changes.

The current state of CI is misleading. The integration test for this PR is actually passing. It's being run in soundness right now because this PR hasn't been rebased since #23 was merged. This means the Github status for this PR is currently showing a pipeline for the integration test that is failing because the Compose "service" the pipeline is trying to run doesn't exist on this branch.

I'll update this branch, which should then at least make the Github checks pass for this PR.

However, the integration test does produce warnings (because there's a conversion of a non-sendable closure to a sendable one), which I imagine would be an error for an adopter if they were running with a stricter compiler flag.

MahdiBM:
Let me investigate this. But for now, the good news is there is no stricter concurrency flag and i doubt there will be one.

If you want to cut a release and do release notes, i can also go the existential-any PR first, so you can do them together if it matters (probably/hopefully that won't need as much time/conversation as this, as it's just some `any`s that the compiler will expect you to explicitly use)

MahdiBM:
So considering the fact that i don't think the integration-tests test-package is public, and also my lack of knowledge about the generator package, i couldn't come to a definite conclusion. But i think:
- It appears the openapi-generator package is generating some non-Sendable stuff? (APIHandler.getGreeting)
- If that's true, then it's a concern and should be resolved.
- But even so, the openapi-generator is set to use up to the "next-minor" version of this package, so merging this PR can't break that.
- The integration tests are running without any strict-concurrency flag. If someone is using the generator with stricter concurrency, their code will likely break honestly IF we don't fix that problem in the openapi-generator before allowing the generator to pull higher minor versions. (I reckon this might have been what @simonjbeaumont meant when saying "I imagine would be an error for an adopter if they were running with a stricter compiler flag")
- Weird thing is the same thing is not happening with the PetstoreConsumerTests 🤔

simonjbeaumont:
Thanks @MahdiBM. I think you've managed to glean what's going on. Let me summarise for you.

We have an `IntegrationTest` Swift package which is colocated in the `swift-openapi-generator` repository. It is a simple package that has dependencies on the generator plugin and the runtime library. 

In CI we clone this package and use `swift package edit` to override the dependency on the package that the PR is for. Using `swift package edit` will override any `upToNextMinor` constraint IIUC. 

So, as expected, for this PR we see the following in the CI logs:

```
[620/621] Compiling Server Server.swift
/code/tmp.run-integration-test.sh.FMusHSr3Pf.noindex/swift-openapi-generator/IntegrationTest/.build/plugins/outputs/integrationtest/Server/OpenAPIGenerator/GeneratedSources/Server.swift:45:31: warning: converting non-sendable function value to '@Sendable (APIHandler) -> ((Operations.getGreeting.Input) async throws -> Operations.getGreeting.Output)' may introduce data races
            using: APIHandler.getGreeting,
```

Depending on the PR this information might be gating or just informational, just like the `api breakage` pipeline.

In this case it just serves to confirm that we need to bump the minor version (it'd be major if we were 1.0, ofc) and folks can expect a potentially breaking change if they choose to upgrade. And hopefully the upgrade will be an explicit action on the adopters part because they'll hopefully be using `upToNextMinor`.

Specifically, on your notes:



> * It appears the openapi-generator package is generating some non-Sendable stuff? (APIHandler.getGreeting)
> * If that's true, then it's a concern and should be resolved.
> * But even so, the openapi-generator is set to use up to the "next-minor" version of this package, so merging this PR can't break that.

Correct. I guess the release flow here would need to be:

1. Merge this PR and cut a 0.2.0 release of runtime.
2. Update generator to use 0.2.0 runtime and determine whether this also needs a minor bump.
3. Also update the integration test.

> * The integration tests are running without any strict-concurrency flag. If someone is using the generator with stricter concurrency, their code will likely break honestly IF we don't fix that problem in the openapi-generator before allowing the generator to pull higher minor versions.
>   
>   * I reckon this might have been what @simonjbeaumont meant when saying "I imagine would be an error for an adopter if they were running with a stricter compiler flag"

Right, that's what I meant. I think we should probably add that to the integration test when we can. Is this something you wanted to chase down as you shepherd this feature through?

MahdiBM:
> I think we should probably add that to the integration test when we can. Is this something you wanted to chase down as you shepherd this feature through?

I think that means updating the Dockerfile with the strict concurrency flag?

simonjbeaumont:
> > I think we should probably add that to the integration test when we can. Is this something you wanted to chase down as you shepherd this feature through?
> 
> I think that means updating the Dockerfile with the strict concurrency flag?

Right. My expectation is, if we were to do that in this PR, we'd see it fail. Might be worth doing to hammer home the nature of this change (at least we'd see that pipeline and the API breakage pipeline agree).

czechboy0 (approved):
lgtm

MahdiBM:
I wonder if this a diagnosis issue by Swift:
```swift
using: APIHandler.updatePet,
/// Warning: Converting non-sendable function value to '@Sendable (APIHandler) -> ((Operations.updatePet.Input) async throws -> Operations.updatePet.Output)' may introduce data races
```
but:
```swift
using: { APIHandler.updatePet($0) },
/// No warnings!
```
Both with `-strict-concurrency=complete`.

docker/Dockerfile
```diff
@@ -19,7 +19,7 @@ RUN echo 'export PATH="$HOME/.tools:$PATH"' >> $HOME/.profile
 # swift-format
 ARG swiftformat_version=508.0.0
 RUN git clone --branch $swiftformat_version --depth 1 https://github.com/apple/swift-format $HOME/.tools/swift-format-source
-RUN cd $HOME/.tools/swift-format-source && swift build -c release
+RUN cd $HOME/.tools/swift-format-source && swift build -c release -Xswiftc -strict-concurrency=complete
```
simonjbeaumont: This is us building `swift-format` for the formatting we do in the CI pipeline. This isn't where we want to be adding this.

This should be added to the `docker-compose.yaml` for the `integration-test` "service".

simonjbeaumont (changes requested):

MahdiBM:
After more investigations, i think it's either a Swift diagnosis error that there are those warnings, or it's a correct warning that we can't do much about anyway, other than working around it like in the comment above.

The function is declared in APIProtocol:
```swift
public protocol APIProtocol: Sendable {
...

func updatePet(_ input: Operations.updatePet.Input) async throws -> Operations.updatePet.Output
}
```
Then UniversalServer uses the function like so:
```swift
using: APIHandler.updatePet,
```
And there is also _nothing_ you can do to make the warning go away, without touching what is in UniversalServer. Even marking everything with `@Sendable` doesn't help (most of the stuff are already Sendable anyway).
I think we can just use the workaround above to suppress the warnings after this PR is merged.

docker/Dockerfile
MahdiBM: yes but i already figured that out and fixed it 😅 this commit is outdated.

simonjbeaumont: Great, and sorry—Github notifications brought me here, didn't spot you'd resolved it already :)

FranzBusch:
I think this is a known bug in Sendable checking right now. Just to be sure, do you mind creating a small reproducer and filing an issue over in the Swift repository

MahdiBM:
I found an existing issue for basically the same behavior, it should be enough considering we now know there is an issue for it: https://github.com/apple/swift/issues/64388

MahdiBM:
Hmm i guess i'll file a new issue considering what we're seeing here and what is in that issue have some differences, although the warning is the same.

simonjbeaumont (approved):

simonjbeaumont:
Thanks @MahdiBM for pushing this over the line!

Discussed this with the wider team and changes to sendable aren't currently considered breaking from a semver perspective. @FranzBusch can you confirm?

PR #24, opened 2023-07-13
Additional unit tests for OpenAPIObjectContainer

rboyce:
### Motivation

Trying to encode an `OpenAPIValueContainer` would previously throw an `EncodingError` if the untyped container contained nested dictionary or array structures. (e.g. `{"dict": {"foo": "bar"}}`) Data could be decoded into this format, but it could not be re-encoded successfully.

### Modifications

When encoding the value of a nested JSON array or object in `OpenAPIValueContainer`, we now handle arrays and dicts by wrapping the inner value in `OpenAPIValueContainer` before passing it to the encoder. 
(note: Based on the rest of the `encode()` implementation, it seems that this may have been expected to happen previously during decoding. Let me know if making changes on the decode side of things is preferable.)

### Result

Additional unit tests added to validate that encoder generates correct output for nested data structures in `OpenAPIValueContainer`, and that these data structures can be decoded and successfully re-endcoded afterward.

### Test Plan

Run `Test_CodableExtensions` tests, see all tests pass.


simonjbeaumont:

Sources/OpenAPIRuntime/Base/OpenAPIValue.swift
```diff
@@ -143,6 +143,18 @@ public struct OpenAPIValueContainer: Codable, Equatable, Hashable, Sendable {
...
+                value.map {
+                    try OpenAPIValueContainer.init(unvalidatedValue: $0)
+                }
+            )
+        case let value as [String: Sendable?]:
+            try container.encode(
+                value.mapValues {
+                    try OpenAPIValueContainer.init(unvalidatedValue: $0)
+                }
+            )
```
simonjbeaumont, line 157: Couple of suggestions here:

- We'll be landing #22 soon which means that you'll need to use existential any, i.e. `any Sendable` in place of `Sendable`, so we may as well get it right here so it doesn't need updating again.
- The rest of the switch statement follows a different style of passing `init(unvalidatedValue:)` to `map` and `mapValues` directly, so could we keep it consistent with these new switch cases?

```suggestion
        case let value as [(any Sendable)?]:
            try container.encode(value.map(OpenAPIValueContainer.init(unvalidatedValue:)))
        case let value as [String: (any Sendable)?]:
            try container.encode(value.mapValues(OpenAPIValueContainer.init(unvalidatedValue:)))
```

simonjbeaumont (approved):
@rboyce Thanks for the patch—and for adding some tests for it.

simonjbeaumont (changes requested):
Oops, meant to mark as "Request changes", with some minor fixups in https://github.com/apple/swift-openapi-runtime/pull/24/files#r1262219996.

czechboy0:
@rboyce Okay I landed the base unit tests - if you'd like, you could port your additional unit tests to the new place (`Test_OpenAPIValue.swift`). 🙏 Thanks

rboyce:
Rebased past #25. This PR now just contains the additional unit tests for mixed nested values; thanks for the fix @czechboy0!

Tests/OpenAPIRuntimeTests/Base/Test_OpenAPIValue.swift
```diff
@@ -193,4 +193,96 @@ final class Test_OpenAPIValue: Test_Runtime {
...
+                    ] as [(any Sendable)?],
+                    "nestedDict": [
+                        "nested": 2
+                    ],
+                ])
+            )
+            let encoder: JSONEncoder = .init()
+            encoder.outputFormatting = [.prettyPrinted, .sortedKeys]
+            let data = try encoder.encode(value)
+            XCTAssertEqual(
```
czechboy0, line 224: Please make it consistent with the other tests by using the utility functions that does the encoding/decoding and equality checks.

Tests/OpenAPIRuntimeTests/Base/Test_OpenAPIValue.swift
```diff
@@ -193,4 +193,96 @@ final class Test_OpenAPIValue: Test_Runtime {
...
     }
+
+    func testEncoding_objectNested_success() throws {
+
+        struct Foo: Encodable {
+            var bar: String
+            var dict: OpenAPIObjectContainer = .init()
+        }
+
+        do {
```
czechboy0, line 204: Why is this extra do scope here?

Tests/OpenAPIRuntimeTests/Base/Test_OpenAPIValue.swift
```diff
@@ -193,4 +193,96 @@ final class Test_OpenAPIValue: Test_Runtime {
...
+    }
+
+    func testDecodeEncodeRoundTrip_objectNested_success() throws {
+
+        struct Foo: Codable {
+            var bar: String
+            var dict: OpenAPIObjectContainer = .init()
+        }
+
+        do {
```
czechboy0, line 255: Same here about the do scope.

czechboy0 (changes requested):

czechboy0:
Thanks, please also update the PR description to reflect the remaining changes.

czechboy0 (approved):
Looks good overall, I'll open another PR bringing it to our style in a bit.

czechboy0:
Fixing up the style: https://github.com/apple/swift-openapi-runtime/pull/33

PR #29, opened 2023-07-27
[Runtime] Multiple content types support

czechboy0:
Runtime changes for https://github.com/apple/swift-openapi-generator/pull/146

### Motivation

See https://github.com/apple/swift-openapi-generator/pull/146 for motivation.

### Modifications

Adds new methods for:
- `extractContentTypeIfPresent` - returns the `content-type` header
- `isValidContentType` - evaluates content type equivalence, including wildcards
- `makeUnexpectedContentTypeError` - returns an error that the generated code can throw

Deprecates this method, which will be removed in a future breaking version:
- `validateContentTypeIfPresent` - only made sense when we supported only 1 content type value for a body

### Result

Enables generated code to support multiple content types.

### Test Plan

Deprecated the tests for the deprecated method, and added unit tests for the new methods.


Tests/OpenAPIRuntimeTests/Deprecated/Test_Deprecated.swift
```diff
@@ -222,6 +222,67 @@ final class Test_Deprecated: Test_Runtime {
         )
     }
 
+    @available(*, deprecated)
+    func testValidateContentType_match() throws {
```
czechboy0, line 226: These were just moved here from Tests/OpenAPIRuntimeTests/Conversion/Test_Converter+Common.swift

Sources/OpenAPIRuntime/Conversion/Converter+Common.swift
```diff
@@ -17,31 +17,68 @@ extension Converter {
 
     // MARK: Miscs
 
-    /// Validates that the Content-Type header field (if present)
```
czechboy0, line 20: One method was removed/moved to Deprecated, and new ones were added. Easier to review in split mode.

gjcairo (approved):

Sources/OpenAPIRuntime/Conversion/MIMEType.swift
```diff
@@ -0,0 +1,156 @@
...
+        let parameters =
+            components
+            .compactMap(parseParameter)
+        self.init(
+            type: typeAndSubtype[0],
+            subtype: typeAndSubtype[1],
+            parameters: Dictionary(
+                parameters,
+                uniquingKeysWith: { a, _ in a }
+            )
```
simonjbeaumont, line 139: Are we being overly permissive here? I skimmed the RFC and it the only thing I can interpret requirements about uniqueness of parameters is this:

> After the media type and subtype names, the remainder
   of the header field is simply a set of parameters, specified in an
   attribute=value notation.  The ordering of parameters is not
   significant.

The use of insignificant order and term "set", I infer uniqueness. Should we be rejecting MIMETypes with duplicates as it can only lead to confusion.

Additionally it makes this not LosslessStringConvertible because a round trip may lose some duplicate parameters.

I think we should be returning `nil` here, WDYT?

Tests/OpenAPIRuntimeTests/Base/Test_MIMEType.swift
```diff
@@ -0,0 +1,102 @@
...
+                        "charset": "UTF-8"
+                    ]
+                ),
+                "APPLICATION/JSON; CHARSET=UTF-8"
+            ),
+
+            // Invalid
+            ("application", nil, nil),
+            ("application/foo/bar", nil, nil),
+            ("", nil, nil),
```
simonjbeaumont, line 94: Is it possible to construct `*/json` currently? If so, should it be possible?

Sources/OpenAPIRuntime/Conversion/Converter+Common.swift
```diff
@@ -17,31 +17,53 @@ extension Converter {
...
-                name: "content-type",
-                as: String.self
-            )
-        else {
-            return
+    ///   - received: The concrete content type to validate against the other.
+    ///   - expected: The expected content type, can contain wildcards.
+    /// - Returns: A Boolean value representing whether the concrete content
+    /// type matches the expected one.
+    public func isValidContentType(received: MIMEType?, expected: String) -> Bool {
```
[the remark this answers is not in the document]
czechboy0: Yeah I don't love the name either, so suggestions welcome. I considered `isMatchingContentType`, `isCompatibleContentType`. It should be a sentence fragment that expects a Bool response, so should start with `is` or `has` or similar.

`validateContentTypeMatches` would make sense if it threw an error and returned Void, but maybe `doesContentTypeMatch`?

Sources/OpenAPIRuntime/Conversion/Converter+Common.swift
```diff
@@ -17,31 +17,53 @@ extension Converter {
...
+        guard let received else {
+            return false
         }
-        guard contentType.localizedCaseInsensitiveContains(substring) else {
-            throw RuntimeError.unexpectedContentTypeHeader(contentType)
+        let receivedContentType = received
+        guard let expectedContentType = MIMEType(expected) else {
+            return false
```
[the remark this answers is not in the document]
czechboy0: > So this is going to return false if the expected input to the function is not a valid mimetype?

Correct.

> Can't we just have this function take a MIMEType for expected?

That would push this decision (initialize from string and handle it returning nil) from this one place to all of the call sites in the generated code, and we'd still have to decide what to do in that scenario, so it was easier to have it take a string and optionally convert here. Do you think we should fail if the expected content type isn't valid? That'd also be reasonable, and we could throw from here, so the callsites wouldn't need to change.

Sources/OpenAPIRuntime/Conversion/MIMEType.swift
```diff
@@ -0,0 +1,156 @@
...
+        case concrete(String)
+
+        public static func == (lhs: Token, rhs: Token) -> Bool {
+            // Case-insensitive
+            lhs.description.lowercased() == rhs.description.lowercased()
+        }
+    }
+
+    /// The type – the first token.
+    public var type: Token
```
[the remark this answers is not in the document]
czechboy0: Yup, I read the RFC as I was writing this, but only parsed the bits that we care about and use. Another reason I made the `MIMEType` type SPI, so folks don't think it's a general MIME type that implements the full RFC. Maybe I should rename it to make that more clear, like `ParsedMIMEType` or something?

Sources/OpenAPIRuntime/Conversion/MIMEType.swift
```diff
@@ -0,0 +1,156 @@
...
+        (["\(type.description)/\(subtype.description)"]
+            + parameters
+            .sorted(by: { a, b in a.key < b.key })
+            .map { "\($0)=\($1)" })
+            .joined(separator: "; ")
+    }
+}
+
+extension String {
+    fileprivate var soar_trimmingLeadingAndTrailingSpaces: Self {
```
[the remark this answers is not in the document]
czechboy0: Oh, I did it because I thought that if our dependencies introduced `trimmingLeadingAndTrailingSpaces`, that it'd break us, but now it's coming back to me that the local module's implementation always wins and doesn't produce a clash error, right?

Sources/OpenAPIRuntime/Conversion/MIMEType.swift
czechboy0: > Additionally it makes this not LosslessStringConvertible because a round trip may lose some duplicate parameters.

That's a great point, but unfortunately there isn't just a `StringConvertible` protocol, which is what I was trying to do here. Should I define our own protocol to avoid the confusion with `Lossless`?

> I think we should be returning nil here, WDYT?

Totally reasonable, I read some advice on this and servers apparently usually keep the first, or the last value, when multiple are found. We can be stricter than that and completely reject (aka ignore) the header if there are multiple, but on the other hand, we don't ever _use_ the parameters, so we don't really care if parameters are duplicated, we only use the type/subtype. Maybe it's more philosophical around inputs in general, whether we should be strict even about inputs we don't use. I'm happy to go either way, with the above, do you still think we should discard the MIME type if it has multiple params with the same name?

Sources/OpenAPIRuntime/Conversion/Converter+Common.swift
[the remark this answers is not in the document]
simonjbeaumont: I'm happy with anything that implies it's testing for loose equivalence, rather than the validity of the first parameter.

I think the sentence fragment argument would apply to a method or computed property. I don't think it _has_ to be here, which is essentially a pure function. I'd probably start with a verb, like `validateContentTypeMatches(actual:expected:)`. But I honestly don't mind at all. Any of your suggestions meet my criteria.

Tests/OpenAPIRuntimeTests/Base/Test_MIMEType.swift
czechboy0: By my reading of https://datatracker.ietf.org/doc/html/rfc2045#section-5.1, that wouldn't be a legal MIME type. We wouldn't reject it at parsing though today, it'd just never positively match any provided MIME type. Should we reject it in parsing you think?

Sources/OpenAPIRuntime/Conversion/Converter+Common.swift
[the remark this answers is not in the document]
czechboy0: Ok, went for `isMatchingContentType`, as that'll extend well for when we start supporting more wildcards as well.

Runtime: https://github.com/apple/swift-openapi-runtime/pull/29/commits/606aa0ee56db8cfee6f61108c4e75a5773571f52
Generator: https://github.com/apple/swift-openapi-generator/pull/146/commits/d31270ab01196f3836d0eca0d34885e3f47b3166

Tests/OpenAPIRuntimeTests/Base/Test_MIMEType.swift
gjcairo: I would argue we probably don't even want to allow creating invalid `MIMETypes`, so making the `init` `throws` and adding a check there would be what I'd go for.

czechboy0: Today we have a failable initializer `init?(_ description: String)` which allows us to reject invalid inputs. We don't try to return _why_ it's invalid, we just return nil if it's invalid.

Maybe the type should reflect the limit in the type system? Like

```swift
struct MIMEType {
  enum Kind {
    case any
    case anySubtype(type: String)
    case concrete(type: String, subtype: String)
  }
  var kind: Kind
  var parameters: [String: String]
}
```

WDYT?

Sources/OpenAPIRuntime/Conversion/MIMEType.swift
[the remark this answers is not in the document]
czechboy0: Fixed: https://github.com/apple/swift-openapi-runtime/pull/29/commits/5752e18b1ae469197ae5e38f7533edafb8ddbc10

Sources/OpenAPIRuntime/Conversion/MIMEType.swift
simonjbeaumont: OK, if you feel like this is conventional then that's fine. I just went to the spec. Maybe it's worth dropping a comment here?

Tests/OpenAPIRuntimeTests/Base/Test_MIMEType.swift
simonjbeaumont: I'm happy with either `init?` or `init throws`. And yes, I think pushing this expressivity into the type system is even better.

But as you say, it's SPI, so I'm not precious over  this.

gjcairo: I'm divided on this - I somewhat like it, but you could still create a `concrete` type with wildcards (or even worse, something invalid like `.concrete(type: "*", subtype: "json")`), so I'm not sure whether the extra complexity is worth it.

czechboy0: I couldn't unsee it, so I redid it, and like this much better: https://github.com/apple/swift-openapi-runtime/pull/29/commits/de2248b6beecc256d2b5be5c74fcd1c849e0f4db

Sources/OpenAPIRuntime/Conversion/MIMEType.swift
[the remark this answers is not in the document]
czechboy0: Went with `OpenAPIMIMEType`, in line with `OpenAPIValue`: https://github.com/apple/swift-openapi-runtime/pull/29/commits/09b68a74d1821fcb694e79f77adf7c05408ae326

Sources/OpenAPIRuntime/Conversion/Converter+Common.swift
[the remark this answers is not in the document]
simonjbeaumont: If we think that it's an invalid use of the function to pass an invalid MIME type to `expected` then I think we should treat it differently from the validation logic, yes. I'd be happy with a thrown error in this case I guess.

Would this be a bug in the generator? If so, maybe we should be adding an assert too?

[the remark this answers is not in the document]
czechboy0: Since this is an SPI function on `Converter`, which are all optimized to make the generated code as short/high signal as possible, I'd be happy to rename the function to reflect the fact that it expects to receive an unvalidated content type string, and be able to handle if it's invalid. Maybe I could rename the parameter label from `expected` to `expectedRaw` or something, and update the doc comments?

And sure, I can make it throw if the expected content type doesn't parse.

This could happen if the exact content type parsing logic in OpenAPIKit, and some more processing in the generator don't exactly match the runtime library's. That could easily happen since generator doesn't depend on an exact, but a floating version of runtime.

Ok, let me make these two changes (rename the label and make it throw on invalid).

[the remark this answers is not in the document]
czechboy0: Ok fixed:
- runtime: https://github.com/apple/swift-openapi-runtime/pull/29/commits/e0dae8558525888f24b2d875375dfd0856b150c6
- generator: https://github.com/apple/swift-openapi-generator/pull/146/commits/248840c47b39f7d633c94b53a584f334e83776c4

Sources/OpenAPIRuntime/Base/OpenAPIMIMEType.swift
```diff
@@ -0,0 +1,170 @@
...
+        )
+        return normalizedLeftParams == normalizedRightParams
+    }
+}
+
+extension OpenAPIMIMEType.Kind: LosslessStringConvertible {
+    public init?(_ description: String) {
+        let typeAndSubtype =
+            description
+            .split(separator: "/")
```
simonjbeaumont, line 88: Nit: could use maxSplits here.

czechboy0: Possibly, but I think that'd complicate us rejecting `foo/bar/baz`, which gets filtered out by the next line where we check exactly for 2 components. We equally want to drop `foo` and `foo/bar/baz`, only accept `foo/bar`.

simonjbeaumont (approved):

Sources/OpenAPIRuntime/Base/OpenAPIMIMEType.swift
simonjbeaumont: OK, fair enough.

PR #37, opened 2023-08-10
[Runtime] Accept header

czechboy0:
### Motivation

SOAR-0003 was accepted, this is the runtime side of the implementation.

### Modifications

- Introduced a new protocol `AcceptableProtocol`, which all the generated, operation-specific Accept enums conform to.
- Introduced a new struct `QualityValue`, which wraps the quality parameter of the content type. Since the precision is capped at 3 decimal places, the internal storage is in 1000's, allowing for more reliable comparisons, as floating point numbers are only used when serialized into headers.
- Introduced a new struct `AcceptHeaderContentType`, generic over `AcceptableProtocol`, which adds `QualityValue` to each generated Accept enum.
- Introduced new extensions on `Converter` that allow setting and getting the Accept header.

### Result

These are the requirements for https://github.com/apple/swift-openapi-generator/pull/185.

### Test Plan

Added unit tests for both `QualityValue` and `AcceptHeaderContentType`, and for the new `Converter` methods.

### DIFF
--- a/Sources/OpenAPIRuntime/Base/Acceptable.swift
+++ b/Sources/OpenAPIRuntime/Base/Acceptable.swift
@@ -13,12 +13,8 @@
 //===----------------------------------------------------------------------===//
 
 /// The protocol that all generated `AcceptableContentType` enums conform to.
-public protocol AcceptableProtocol: RawRepresentable, Sendable, Equatable, Hashable where RawValue == String {
-
-    /// Returns the default set of acceptable content types for this type, in
-    /// the order specified in the OpenAPI document.
-    static var defaultValues: [Self] { get }
-}
+public protocol AcceptableProtocol: RawRepresentable, Sendable, Equatable, Hashable, CaseIterable
+where RawValue == String {}
 
 /// A quality value used to describe the order of priority in a comma-separated
 /// list of values, such as in the Accept header.
@@ -101,7 +97,7 @@ extension Array {
     /// Returns the default values for the acceptable type.
     public static func defaultValues<T: AcceptableProtocol>() -> [AcceptHeaderContentType<T>]
     where Element == AcceptHeaderContentType<T> {
-        T.defaultValues.map { .init(contentType: $0) }
+        T.allCases.map { .init(contentType: $0) }
     }
 }
 
@@ -135,7 +131,7 @@ public struct AcceptHeaderContentType<ContentType: AcceptableProtocol>: Sendable
     /// Returns the default set of acceptable content types for this type, in
     /// the order specified in the OpenAPI document.
     public static var defaultValues: [Self] {
-        ContentType.defaultValues.map { .init(contentType: $0) }
+        ContentType.allCases.map { .init(contentType: $0) }
     }
 }
 
--- a/Sources/OpenAPIRuntime/Base/Acceptable.swift
+++ b/Sources/OpenAPIRuntime/Base/Acceptable.swift
@@ -1,0 +1,160 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// The protocol that all generated `AcceptableContentType` enums conform to.
+public protocol AcceptableProtocol: RawRepresentable, Sendable, Equatable, Hashable where RawValue == String {
+
+    /// Returns the default set of acceptable content types for this type, in
+    /// the order specified in the OpenAPI document.
+    static var defaultValues: [Self] { get }
+}
+
+/// A quality value used to describe the order of priority in a comma-separated
+/// list of values, such as in the Accept header.
+public struct QualityValue: Sendable, Equatable, Hashable {
+
+    /// As the quality value only retains up to and including 3 decimal digits,
+    /// we store it in terms of the thousands.
+    ///
+    /// This allows predictable equality comparisons and sorting.
+    ///
+    /// For example, 1000 thousands is the quality value of 1.0.
+    private let thousands: UInt16
+
+    /// Creates a new quality value of the default value 1.0.
+    public init() {
+        self.thousands = 1000
+    }
+
+    /// Returns a Boolean value indicating whether the quality value is
+    /// at its default value 1.0.
+    public var isDefault: Bool {
+        thousands == 1000
+    }
+
+    /// Creates a new quality value from the provided floating-point number.
+    ///
+    /// - Precondition: The value must be between 0.0 and 1.0, inclusive.
+    public init(doubleValue: Double) {
+        precondition(
+            doubleValue >= 0.0 && doubleValue <= 1.0,
+            "Input number into quality value is out of range"
+        )
+        self.thousands = UInt16(doubleValue * 1000)
+    }
+
+    /// The value represented as a floating-point number between 0.0 and 1.0, inclusive.
+    public var doubleValue: Double {
+        Double(thousands) / 1000
+    }
+}
+
+extension QualityValue: RawRepresentable {
+    public init?(rawValue: String) {
+        guard let doubleValue = Double(rawValue) else {
+            return nil
+        }
+        self.init(doubleValue: doubleValue)
+    }
+
+    public var rawValue: String {
+        String(format: "%0.3f", doubleValue)
+    }
+}
+
+extension QualityValue: Comparable {
+    public static func < (lhs: QualityValue, rhs: QualityValue) -> Bool {
+        lhs.thousands > rhs.thousands
+    }
+}
+
+extension QualityValue: ExpressibleByIntegerLiteral {
+    public init(integerLiteral value: UInt16) {
+        self.thousands = value * 1000
+    }
+}
+
+extension QualityValue: ExpressibleByFloatLiteral {
+    public init(floatLiteral value: Double) {
+        self.init(doubleValue: value)
+    }
+}
+
+/// A wrapper of an individual content type in the accept header.
+public struct AcceptHeaderContentType<T: AcceptableProtocol>: Sendable, Equatable, Hashable {
+
+    /// The quality value of this content type.
+    ///
+    /// Used to describe the order of priority in a comma-separated
+    /// list of values.
+    ///
+    /// Content types with a higher priority should be preferred by the server
+    /// when deciding which content type to use in the response.
+    ///
+    /// Also called the "q-factor" or "q-value".
+    public let quality: QualityValue
+
+    /// The value representing the content type.
+    public let contentType: T
+
+    /// Creates a new content type from the provided parameters.
+    /// - Parameters:
+    ///   - quality: The quality of the content type, between 0.0 and 1.0.
+    ///   - value: The value representing the content type.
+    /// - Precondition: Priority must be in the range 0.0 and 1.0 inclusive.
+    public init(quality: QualityValue = 1.0, contentType: T) {
+        self.quality = quality
+        self.contentType = contentType
+    }
+
+    /// Returns the default set of acceptable content types for this type, in
+    /// the order specified in the OpenAPI document.
+    public static var defaultValues: [Self] {
+        T.defaultValues.map { .init(contentType: $0) }
+    }
+}
+
+extension AcceptHeaderContentType: RawRepresentable {
+    public init?(rawValue: String) {
+        guard let validMimeType = OpenAPIMIMEType(rawValue) else {
+            // Invalid MIME type.
+            return nil
+        }
+        let quality: QualityValue
+        if let rawQuality = validMimeType.parameters["q"] {
+            guard let parsedQuality = QualityValue(rawValue: rawQuality) else {
+                // Invalid quality parameter.
+                return nil
+            }
+            quality = parsedQuality
+        } else {
+            quality = 1.0
+        }
+        guard let typeAndSubtype = T.init(rawValue: validMimeType.kind.description.lowercased()) else {
+            // Invalid type/subtype.
+            return nil
+        }
+        self.init(quality: quality, contentType: typeAndSubtype)
+    }
+
+    public var rawValue: String {
+        contentType.rawValue + (quality.isDefault ? "" : "; q=\(quality.rawValue)")
+    }
+}
+
+extension AcceptHeaderContentType: Comparable {
+    public static func < (lhs: AcceptHeaderContentType<T>, rhs: AcceptHeaderContentType<T>) -> Bool {
+        lhs.quality < rhs.quality
+    }
+}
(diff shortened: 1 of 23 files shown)


Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    public var rawValue: String {
+        String(format: "%0.3f", doubleValue)
+    }
+}
+
+extension QualityValue: Comparable {
+    public static func < (lhs: QualityValue, rhs: QualityValue) -> Bool {
+        lhs.thousands > rhs.thousands
```
glbrntt, line 78: Is this intentionally backwards?

Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,160 @@
...
+    /// list of values.
+    ///
+    /// Content types with a higher priority should be preferred by the server
+    /// when deciding which content type to use in the response.
+    ///
+    /// Also called the "q-factor" or "q-value".
+    public let quality: QualityValue
+
+    /// The value representing the content type.
+    public let contentType: T
```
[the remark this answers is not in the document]
czechboy0: No, I'll change it. I had to make `QualityValue` have `let`s because we run validation on creation, and must have accidentally changed it here as well.

Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,160 @@
...
+
+    /// Creates a new content type from the provided parameters.
+    /// - Parameters:
+    ///   - quality: The quality of the content type, between 0.0 and 1.0.
+    ///   - value: The value representing the content type.
+    /// - Precondition: Priority must be in the range 0.0 and 1.0 inclusive.
+    public init(quality: QualityValue = 1.0, contentType: T) {
```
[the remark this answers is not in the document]
czechboy0: Yup will change.

Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,160 @@
...
+        if let rawQuality = validMimeType.parameters["q"] {
+            guard let parsedQuality = QualityValue(rawValue: rawQuality) else {
+                // Invalid quality parameter.
+                return nil
+            }
+            quality = parsedQuality
+        } else {
+            quality = 1.0
+        }
+        guard let typeAndSubtype = T.init(rawValue: validMimeType.kind.description.lowercased()) else {
```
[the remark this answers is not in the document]
czechboy0: Sure, I don't know if I'm misremembering but I thought there were cases when using the generic parameter to init required the explicit `.init`. If it compiles without, I'll change it.

Sources/OpenAPIRuntime/Base/Acceptable.swift
czechboy0: That's fair, yeah I'll flip it.

Sources/OpenAPIRuntime/Base/Acceptable.swift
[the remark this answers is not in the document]
czechboy0: Ok fixed.

Sources/OpenAPIRuntime/Base/Acceptable.swift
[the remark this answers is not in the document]
czechboy0: Fixed.

Sources/OpenAPIRuntime/Base/Acceptable.swift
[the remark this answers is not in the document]
czechboy0: Fixed.

Sources/OpenAPIRuntime/Base/Acceptable.swift
czechboy0: I removed the conformance on Comparable, it was too confusing. I'll create focused methods instead.

Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,177 @@
...
+
+extension Array where Element == QualityValue {
+
+    /// Returns a sorted array of quality values, where the highest
+    /// priority items come first.
+    public func sortedByQuality() -> Self {
+        sorted { a, b in
+            a.doubleValue < b.doubleValue
+        }
+    }
```
glbrntt, line 96: Presumably this should be `a.doubleValue > b.doubleValue`?

Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,177 @@
...
+        self.thousands = UInt16(doubleValue * 1000)
+    }
+
+    /// The value represented as a floating-point number between 0.0 and 1.0, inclusive.
+    public var doubleValue: Double {
+        Double(thousands) / 1000
+    }
+}
+
+extension QualityValue: RawRepresentable {
```
glbrntt, line 63: What do we get from this being `RawRepresentable`? It feels a bit odd having the `RawValue` be a `String`; on first though `Double` seems like a more natural fit.

Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,177 @@
...
+extension Array where Element: AcceptableProtocol {
+
+    /// Returns the default values for the acceptable type.
+    public var defaultValues: Self {
+        Element.defaultValues
+    }
+}
+
+/// A wrapper of an individual content type in the accept header.
+public struct AcceptHeaderContentType<T: AcceptableProtocol>: Sendable, Equatable, Hashable {
```
glbrntt, line 108: Can we use `ContentType` instead of `T`?

Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    public var rawValue: String {
+        contentType.rawValue + (quality.isDefault ? "" : "; q=\(quality.rawValue)")
+    }
+}
+
+extension AcceptHeaderContentType {
+
+    // TODO: Can we spell this as an extension of an array?
```
glbrntt, line 171: I think the spelling would be something like:

```swift
extension Array {
  func sortedByQuality<T: AcceptableProtocol>() -> [AcceptHeaderContentType<T>] where Element == AcceptHeaderContentType<T> { 
    // ...
  } 
}


```

Sources/OpenAPIRuntime/Base/Acceptable.swift
czechboy0: Yes, thank you. 🤦‍♂️

Sources/OpenAPIRuntime/Base/Acceptable.swift
czechboy0: That's how we convert from/to raw parameter values. It feels cleaner than having a specially named `init?(fromHeaderValueString:)` and `var headerValueString: String { ... }`.

Sources/OpenAPIRuntime/Base/Acceptable.swift
czechboy0: Yup!

Sources/OpenAPIRuntime/Base/Acceptable.swift
czechboy0: Ah that's what I was looking for – thanks! 🙏

Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    public var rawValue: String {
+        String(format: "%0.3f", doubleValue)
+    }
+}
+
+extension QualityValue: ExpressibleByIntegerLiteral {
+    public init(integerLiteral value: UInt16) {
+        self.thousands = value * 1000
```
gjcairo, line 74: Feels like we should also precondition here that we're within a valid range - and multiplying by 1000 is probably unwanted too.

Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,164 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// The protocol that all generated `AcceptableContentType` enums conform to.
+public protocol AcceptableProtocol: RawRepresentable, Sendable, Equatable, Hashable, CaseIterable
```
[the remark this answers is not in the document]
czechboy0: Let's fix it all in one go later: https://github.com/apple/swift-openapi-generator/issues/210

Sources/OpenAPIRuntime/Base/Acceptable.swift
czechboy0: Good catch - added a precondition.

Multiplying it is correct, as the internal storage is in thousands, so 1000 maps to quality value 1.

czechboy0: https://github.com/apple/swift-openapi-runtime/pull/37/commits/15d2b021ec6f88c5d845a3d59371311ac44de2cd

gjcairo: Ah okay, I kinda assumed that if using this init, we would be passing a `thousands` value and not actually a quality value, although I understand that's confusing and wrong. So essentially the only allowed values here are 0 or 1, right?

czechboy0: Yup.

Tests/OpenAPIRuntimeTests/Base/Test_Acceptable.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+}
+
+final class Test_QualityValue: Test_Runtime {
+    func test() {
+        XCTAssertEqual(QualityValue().doubleValue, 1.0)
+        XCTAssertTrue(QualityValue().isDefault)
+        XCTAssertFalse(QualityValue(doubleValue: 0.5).isDefault)
+        XCTAssertEqual(QualityValue(doubleValue: 0.5).doubleValue, 0.5)
+        XCTAssertEqual(QualityValue(floatLiteral: 0.5).doubleValue, 0.5)
```
gjcairo, line 99: I think it would be worth adding an assertion/a test for creating a `QualityValue` from a number with 3 decimal non-zero digits, to make sure that we're preserving them properly (which is basically the whole reason why we store it as `thousands` internally)

Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,167 @@
...
+        }
+        self.init(doubleValue: doubleValue)
+    }
+
+    public var rawValue: String {
+        String(format: "%0.3f", doubleValue)
+    }
+}
+
+extension QualityValue: ExpressibleByIntegerLiteral {
```
glbrntt, line 71: Is this actually necessary? I assume the compiler would infer `0`/`1` to be a `Float` if this didn't exist and use the `ExpressibleByFloatLiteral` conformance.

Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,167 @@
...
+    /// As the quality value only retains up to and including 3 decimal digits,
+    /// we store it in terms of the thousands.
+    ///
+    /// This allows predictable equality comparisons and sorting.
+    ///
+    /// For example, 1000 thousands is the quality value of 1.0.
+    private let thousands: UInt16
+
+    /// Creates a new quality value of the default value 1.0.
+    public init() {
```
glbrntt, line 31: Without reading the docs it's not obvious that `let q = QualityValue()` would default to 1.0. From a readability perspective it might make more sense to not have a default init.

Sources/OpenAPIRuntime/Base/Acceptable.swift
```diff
@@ -0,0 +1,167 @@
...
+    /// when deciding which content type to use in the response.
+    ///
+    /// Also called the "q-factor" or "q-value".
+    public var quality: QualityValue
+
+    /// Creates a new content type from the provided parameters.
+    /// - Parameters:
+    ///   - value: The value representing the content type.
+    ///   - quality: The quality of the content type, between 0.0 and 1.0.
+    /// - Precondition: Priority must be in the range 0.0 and 1.0 inclusive.
```
glbrntt, line 117:
```suggestion
    /// - Precondition: Quality must be in the range 0.0 and 1.0 inclusive.
```

Sources/OpenAPIRuntime/Base/Acceptable.swift
czechboy0: I mean, I personally try to avoid jumping from int -> float -> int (internal representation), for the fear of some tiny inaccuracy sneaking in. With this initializer, we go from int -> int (internal representation of input * 1000).

Sources/OpenAPIRuntime/Base/Acceptable.swift
czechboy0: It is the default according to the RFC. Are you saying the user should be forced to always specify quality value, even if they don't want to customize it? That might be strange, as `application/json, text/plain` and `application/json; q=1.0, text/plain; q=1.0` are completely equivalent, and unless people customize the quality value to be anything other than 1.0, they usually don't explicitly specify it.

Sources/OpenAPIRuntime/Base/Acceptable.swift
czechboy0: Good catch, thanks. I started calling them "priority" originally but then (clearly imperfectly) updated the codebase.

Tests/OpenAPIRuntimeTests/Base/Test_Acceptable.swift
czechboy0: Will do.

Sources/OpenAPIRuntime/Base/Acceptable.swift
glbrntt: Fair enough. IMO I don't think it's worth it for exactly two values. Not a big deal though!

Sources/OpenAPIRuntime/Base/Acceptable.swift
glbrntt: No, I would just expect it to be defaulted at a higher level like how it is for `AcceptHeaderContentType`.

czechboy0: Oh sure, turns out we're only using it in tests. I'll remove it.

czechboy0: Removed in https://github.com/apple/swift-openapi-runtime/pull/37/commits/068b34e4e9da4faa920e692f46ab1b77fcc69cef

Tests/OpenAPIRuntimeTests/Base/Test_Acceptable.swift
czechboy0: Fixed in https://github.com/apple/swift-openapi-runtime/pull/37/commits/068b34e4e9da4faa920e692f46ab1b77fcc69cef

gjcairo (approved):

PR #39, opened 2023-08-14
add converter methods for url forms

bfrearson:
### Motivation

[Linked generator PR](https://github.com/apple/swift-openapi-generator/pull/195)

This is a draft PR to receive feedback on [issue 182](https://github.com/apple/swift-openapi-generator/issues/182). However the latest discussion there recommends that we work towards a custom URLFormEncoder/Decoder that should replace the majority of the logic in these methods. Ideally, they would look identical to their JSON method counterparts: `encoder.encode(value)`, rather that implementing the steps here.

### Modifications

Add converter methods for a new url form method. For this implementation, we're leaning on the JSONEncoder, which definitely has its drawbacks. As the discussion on the issue evolves, it's clear that a more custom approach would be better.

### Result

The matching [generator PR](https://github.com/apple/swift-openapi-generator/pull/195) changes the implementation of url encoded forms, but the runtime changes here are additional methods, so nothing will change with existing implementations

### Test Plan

No tests added as of yet.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/CurrencyExtensions.swift
+++ b/Sources/OpenAPIRuntime/Conversion/CurrencyExtensions.swift
@@ -176,11 +176,41 @@ extension Converter {
     ) throws -> T {
         try decoder.decode(T.self, from: data)
     }
+    
+    func convertURLEncodedFormToCodable<T:Decodable>(
+        _ data: Data
+    ) throws -> T {
+        
+        let decoder = URIDecoder(configuration: .init(style: .form,
+                                                      explode: true,
+                                                      spaceEscapingCharacter: .percentEncoded,
+                                                      dateTranscoder: configuration.dateTranscoder))
+        guard let uriString = String(data: data, encoding: .utf8) else {
+            throw RuntimeError.failedToSerializeCodableData
+        }
+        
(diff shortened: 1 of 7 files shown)


Sources/OpenAPIRuntime/Conversion/CurrencyExtensions.swift
```diff
@@ -158,12 +158,56 @@ extension Converter {
...
+        let pairs = urlFormString.split(separator: "&")
+        for pair in pairs {
+            let components = pair.split(separator: "=")
+            let key = String(components[0])
+            let value = String(components[1])
+            dictionary[key] = value
+        }
+        
+        //deserialize into JSON representation
+        let jsonData = try JSONSerialization.data(withJSONObject: dictionary)
```
czechboy0, line 181: What happens if the Codable type has a property `foo: Int`? I suspect it won't parse, as in the JSON it'll be `"1"` instead of `1`? Same for other non-string primitive types that have a native JSON type?

bfrearson: Yes, good spot - we certainly need some unit tests to cover non trivial conversions here, and this implementation has a fair few limitations!

PR #41, opened 2023-08-21
Introduce URIEncoder and URIDecoder types

czechboy0:
### Motivation

Fixes https://github.com/apple/swift-openapi-generator/issues/192.

For refactoring how we encode:
- path parameters
- query parameters
- headers
- (new feature, coming later) `application/x-www-form-urlencoded` bodies

Supports:
- form + explode
- form + unexplode
- simple + explode
- simple + unexplode
- form where space is encoded as + instead of %20 (for bodies) + explode
- form where space is encoded as + instead of %20 (for bodies) + unexplode

### Modifications

First step - introduce two new types: `URIEncoder` and `URIDecoder`.

They're configurable types that can handle the URI Template (RFC 6570) form and simple styles, refined by OpenAPI 3.0.3, and also the `application/x-www-form-urlencoded` format (mostly identical to URI Template).

### Result

The types can be used now, subsequent PRs will integrate them.

### Test Plan

Added unit tests for the individual parts of the coder, but also roundtrip tests.


bfrearson:
This looks great!

I think it's flexible enough to use wherever.

Do you think we should support percent encoding from the beginning? I can see a lot of things breaking if you included a literal `=` or `&` in the input.

czechboy0:
@bfrearson Ok so I got it all working end to end now, without nesting (that's a future enhancement) - WDYT? This isn't polished yet, but tests are passing.

Tests/OpenAPIRuntimeTests/URICoder/Test_URICodingRoundtrip.swift
```diff
@@ -0,0 +1,240 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import XCTest
+@testable import OpenAPIRuntime
+
+final class Test_URICodingRoundtrip: Test_Runtime {
+
```
czechboy0, line 18: TODO

Sources/OpenAPIRuntime/URICoder/Common/URICoderConfiguration.swift
```diff
@@ -0,0 +1,72 @@
...
+    private init(style: Style, explode: Bool, spaceEscapingCharacter: String) {
+        self.style = style
+        self.explode = explode
+        self.spaceEscapingCharacter = spaceEscapingCharacter
+    }
+
+    static let formExplode: Self = .init(
+        style: .form,
+        explode: true,
+        spaceEscapingCharacter: "%20"
```
bfrearson, line 40: Minor point, but spaceEscapingCharacter might be better as an enum so we can be sure that we're always using the same value:

```
enum SpaceEscapingCharacter: String {
    case percent = "%20"
    case plus = "+"
}
```

Sources/OpenAPIRuntime/URICoder/Decoding/URIDecoder.swift
```diff
@@ -0,0 +1,99 @@
...
+
+    /// Make multiple decode calls on the parsed URI.
+    ///
+    /// Use to avoid repeatedly reparsing the raw string.
+    /// - Parameters:
+    ///   - data: The URI-encoded string.
+    ///   - calls: The closure that contains 0 or more calls to
+    ///     URICachedDecoder's decode method.
+    /// - Returns: The result of the closure invocation.
+    func withCachedParser<R>(
```
bfrearson, line 61: I'd be interested how much of a performance improvement this offers! Obviously optimisation is a good thing, but just curious if it makes a meaningful impact.

bfrearson (changes requested):
Looks really good - just that minor suggestion about using an enum for space encoding

Sources/OpenAPIRuntime/URICoder/Common/URICoderConfiguration.swift
czechboy0: Sure, it's not like there'll be a third case any time soon. I'll change it.

Sources/OpenAPIRuntime/URICoder/Decoding/URIDecoder.swift
czechboy0: Not sure, I was worried about e.g. 10 key-value pairs, and we'd have to parse the string 10 times when pulling each individual value out.

This way, we at least cache the parsing, only the decoding is done from scratch every time, which I think is fine, as it doesn't scale up as much with the length of the encoded string (unlike parsing).

Sources/OpenAPIRuntime/URICoder/Common/URICoderConfiguration.swift
czechboy0: https://github.com/apple/swift-openapi-runtime/pull/41/commits/a24c805046b2930af0ac919fd44f23f67dc39bb8

glbrntt (approved):
Looks great!

Sources/OpenAPIRuntime/URICoder/Common/URIEncodedNode.swift
```diff
@@ -0,0 +1,145 @@
...
+            precondition(
+                intValue == array.count,
+                "Unkeyed container inserting at an incorrect index"
+            )
+            self = .unset
+            array.append(childValue)
+            self = .array(array)
+        case .unset:
+            if let _ = key.intValue {
+                self = .array([childValue])
```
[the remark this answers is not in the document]
czechboy0: Sure, I'll add that to be extra sure there isn't a programmer error hiding in there.

PR #46, opened 2023-09-01
[WIP] Async body currency type

czechboy0:
### Motivation

An early prototype of #9.

### Modifications

TODO

### Result

TODO

### Test Plan

TODO

### DIFF
--- a/Sources/OpenAPIRuntime/Interface/Body.swift
+++ b/Sources/OpenAPIRuntime/Interface/Body.swift
@@ -1,0 +1,535 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import class Foundation.NSLock
+import protocol Foundation.LocalizedError
+import struct Foundation.Data  // only for convenience initializers
+
+/// The type representing a request or response body.
+public final class Body: @unchecked Sendable {
+
+    /// The underlying data type.
+    public typealias DataType = ArraySlice<UInt8>
+
+    /// How many times the provided sequence can be iterated.
+    public enum IterationBehavior: Sendable {
+
+        /// The input sequence can only be iterated once.
+        ///
+        /// If a retry or a redirect is encountered, fail the call with
+        /// a descriptive error.
+        case single
+
+        /// The input sequence can be iterated multiple times.
+        ///
+        /// Supports retries and redirects, as a new iterator is created each
+        /// time.
+        case multiple
+    }
+
+    /// How many times the provided sequence can be iterated.
+    public let iterationBehavior: IterationBehavior
+
+    /// The total length of the body, if known.
+    public enum Length: Sendable {
+
+        /// Total length not known yet.
+        case unknown
+
+        /// Total length is known.
+        case known(Int)
+    }
+
+    /// The total length of the body, if known.
+    public let length: Length
+
+    /// The underlying type-erased async sequence.
+    private let sequence: BodySequence
+
+    /// A lock for shared mutable state.
+    private let lock: NSLock = {
+        let lock = NSLock()
+        lock.name = "com.apple.swift-openapi-generator.runtime.body"
+        return lock
+    }()
+
+    /// Whether an iterator has already been created.
+    private var locked_iteratorCreated: Bool = false
+
+    private init(
+        sequence: BodySequence,
+        length: Length,
+        iterationBehavior: IterationBehavior
+    ) {
+        self.sequence = sequence
+        self.length = length
+        self.iterationBehavior = iterationBehavior
+    }
+}
+
+extension Body: Equatable {
+    public static func == (
+        lhs: Body,
+        rhs: Body
+    ) -> Bool {
+        ObjectIdentifier(lhs) == ObjectIdentifier(rhs)
+    }
+}
+
+extension Body: Hashable {
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(ObjectIdentifier(self))
+    }
+}
+
+// MARK: - Creating the Body
+
+extension Body {
+
+    public convenience init(
+        data: DataType,
+        length: Length
+    ) {
+        self.init(
+            dataChunks: [data],
+            length: length
+        )
+    }
+
+    public convenience init(
+        data: DataType
+    ) {
+        self.init(
+            dataChunks: [data],
+            length: .known(data.count)
+        )
+    }
+
+    public convenience init<S: Sequence>(
+        dataChunks: S,
+        length: Length,
+        iterationBehavior: IterationBehavior
+    ) where S.Element == DataType {
+        self.init(
+            sequence: .init(WrappedSyncSequence(sequence: dataChunks)),
+            length: length,
+            iterationBehavior: iterationBehavior
+        )
+    }
+
+    public convenience init<C: Collection>(
+        dataChunks: C,
+        length: Length
+    ) where C.Element == DataType {
+        self.init(
+            sequence: .init(WrappedSyncSequence(sequence: dataChunks)),
+            length: length,
+            iterationBehavior: .multiple
+        )
+    }
+
+    public convenience init<C: Collection>(
+        dataChunks: C
+    ) where C.Element == DataType {
+        self.init(
+            sequence: .init(WrappedSyncSequence(sequence: dataChunks)),
+            length: .known(dataChunks.map(\.count).reduce(0, +)),
+            iterationBehavior: .multiple
+        )
+    }
+
+    public convenience init(
+        stream: AsyncThrowingStream<DataType, any Error>,
+        length: Body.Length
+    ) {
+        self.init(
+            sequence: .init(stream),
+            length: length,
+            iterationBehavior: .single
+        )
+    }
+
+    public convenience init(
+        stream: AsyncStream<DataType>,
+        length: Body.Length
+    ) {
+        self.init(
+            sequence: .init(stream),
+            length: length,
+            iterationBehavior: .single
+        )
+    }
+
+    public convenience init<S: AsyncSequence>(
+        sequence: S,
+        length: Body.Length,
+        iterationBehavior: IterationBehavior
+    ) where S.Element == DataType {
+        self.init(
+            sequence: .init(sequence),
+            length: length,
+            iterationBehavior: iterationBehavior
+        )
+    }
+}
+
+// MARK: - Consuming the body
+
+extension Body: AsyncSequence {
+    public typealias Element = DataType
+    public typealias AsyncIterator = Iterator
+    public func makeAsyncIterator() -> AsyncIterator {
+        if iterationBehavior == .single {
+            lock.lock()
+            defer {
+                lock.unlock()
+            }
+            guard !locked_iteratorCreated else {
+                fatalError(
+                    "OpenAPIRuntime.Body attempted to create a second iterator, but the underlying sequence is only safe to be iterated once."
+                )
+            }
+            locked_iteratorCreated = true
+        }
+        return sequence.makeAsyncIterator()
+    }
+}
+
+// MARK: - Transforming the body
+
+extension Body {
+
+    /// Creates a body where each chunk is transformed by the provided closure.
+    /// - Parameter transform: A mapping closure.
+    /// - Throws: If a known length was provided to this body at
+    /// creation time, the transform closure must not change the length of
+    /// each chunk.
+    public func mapChunks(
+        _ transform: @escaping @Sendable (Element) async -> Element
+    ) -> Body {
+        let validatedTransform: @Sendable (Element) async -> Element
+        switch length {
+        case .known:
+            validatedTransform = { element in
+                let transformedElement = await transform(element)
+                guard transformedElement.count == element.count else {
+                    fatalError(
+                        "OpenAPIRuntime.Body.mapChunks transform closure attempted to change the length of a chunk in a body which has a total length specified, this is not allowed."
+                    )
+                }
+                return transformedElement
+            }
+        case .unknown:
+            validatedTransform = transform
+        }
+        return Body(
+            sequence: map(validatedTransform),
+            length: length,
+            iterationBehavior: iterationBehavior
+        )
+    }
+}
+
+// MARK: - Consumption utils
+
+extension Body {
+
+    /// An error thrown by the `collect` function when the body contains more
+    /// than the maximum allowed number of bytes.
+    private struct TooManyBytesError: Error, CustomStringConvertible, LocalizedError {
+        let maxBytes: Int
+
+        var description: String {
+            "OpenAPIRuntime.Body contains more than the maximum allowed \(maxBytes) bytes."
+        }
+
+        var errorDescription: String? {
+            description
+        }
+    }
+
+    /// An error thrown by the `collect` function when another iteration of
+    /// the body is not allowed.
+    private struct TooManyIterationsError: Error, CustomStringConvertible, LocalizedError {
+
+        var description: String {
+            "OpenAPIRuntime.Body attempted to create a second iterator, but the underlying sequence is only safe to be iterated once."
+        }
+
+        var errorDescription: String? {
+            description
+        }
+    }
+
+    /// Accumulates the full body in-memory into a single buffer
+    /// up to `maxBytes` and returns it.
+    /// - Parameters:
+    ///   - maxBytes: The maximum number of bytes this method is allowed
+    ///     to accumulate in memory before it throws an error.
+    /// - Throws: `TooManyBytesError` if the the sequence contains more
+    ///   than `maxBytes`.
+    public func collect(upTo maxBytes: Int) async throws -> DataType {
+
+        // As a courtesy, check if another iteration is allowed, and throw
+        // an error instead of fatalError here if the user is trying to
+        // iterate a sequence for the second time, if it's only safe to be
+        // iterated once.
+        if iterationBehavior == .single {
+            try {
+                lock.lock()
+                defer {
+                    lock.unlock()
+                }
+                guard !locked_iteratorCreated else {
+                    throw TooManyIterationsError()
+                }
+            }()
+        }
+
+        var buffer = DataType.init()
+        for try await chunk in self {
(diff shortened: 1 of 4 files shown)


Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import class Foundation.NSLock
+import protocol Foundation.LocalizedError
+import struct Foundation.Data  // only for convenience initializers
+
+/// The type representing a request or response body.
+public final class Body: @unchecked Sendable {
```
FranzBusch, line 20: Why is this a class? Also can we get rid of the `@unchecked Sendable` here and rather make ourselves a `LockedValueBox` abstraction that is conditionally `Sendable` when its contents are. This way we are making sure we really are `Sendable`

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+
+    /// The underlying data type.
+    public typealias DataType = ArraySlice<UInt8>
+
+    /// How many times the provided sequence can be iterated.
+    public enum IterationBehavior: Sendable {
```
FranzBusch, line 26: This probably should be a `struct`.

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+        ///
+        /// If a retry or a redirect is encountered, fail the call with
+        /// a descriptive error.
+        case single
+
+        /// The input sequence can be iterated multiple times.
+        ///
+        /// Supports retries and redirects, as a new iterator is created each
+        /// time.
+        case multiple
```
FranzBusch, line 38: Are we ever expecting an `AsyncSequence` based body to be `multiple` here? Right now that just doesn't exist in the ecosystem.

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    /// How many times the provided sequence can be iterated.
+    public let iterationBehavior: IterationBehavior
+
+    /// The total length of the body, if known.
+    public enum Length: Sendable {
```
FranzBusch, line 45: Same here. Probably a `struct`

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+        length: Length,
+        iterationBehavior: IterationBehavior
+    ) {
+        self.sequence = sequence
+        self.length = length
+        self.iterationBehavior = iterationBehavior
+    }
+}
+
+extension Body: Equatable {
```
FranzBusch, line 81: I am not sure if we really need `Equatable` and `Hashable` on the body

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+        stream: AsyncStream<DataType>,
+        length: Body.Length
+    ) {
+        self.init(
+            sequence: .init(stream),
+            length: length,
+            iterationBehavior: .single
+        )
+    }
+
```
FranzBusch, line 173: These two methods seem unnecessary when you already have the generic one which takes an `AsyncSequence`

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+                lock.unlock()
+            }
+            guard !locked_iteratorCreated else {
+                fatalError(
+                    "OpenAPIRuntime.Body attempted to create a second iterator, but the underlying sequence is only safe to be iterated once."
+                )
+            }
+            locked_iteratorCreated = true
+        }
+        return sequence.makeAsyncIterator()
```
FranzBusch, line 205: We shouldn't have to do this at all. If we just delegate to call `makeAsyncIterator` on the underlying sequence then that base async sequence should enforce if it is `unicast` or `multicast`.

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+// MARK: - Transforming the body
+
+extension Body {
+
+    /// Creates a body where each chunk is transformed by the provided closure.
+    /// - Parameter transform: A mapping closure.
+    /// - Throws: If a known length was provided to this body at
+    /// creation time, the transform closure must not change the length of
+    /// each chunk.
+    public func mapChunks(
```
FranzBusch, line 218: Interesting. What is the use-case for this?

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+            "OpenAPIRuntime.Body attempted to create a second iterator, but the underlying sequence is only safe to be iterated once."
+        }
+
+        var errorDescription: String? {
+            description
+        }
+    }
+
+    /// Accumulates the full body in-memory into a single buffer
+    /// up to `maxBytes` and returns it.
```
FranzBusch, line 276: This contradicts each other. Collects the whole body and up to.

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+    ///     to accumulate in memory before it throws an error.
+    /// - Throws: `TooManyBytesError` if the the sequence contains more
+    ///   than `maxBytes`.
+    public func collect(upTo maxBytes: Int) async throws -> DataType {
+
+        // As a courtesy, check if another iteration is allowed, and throw
+        // an error instead of fatalError here if the user is trying to
+        // iterate a sequence for the second time, if it's only safe to be
+        // iterated once.
+        if iterationBehavior == .single {
```
FranzBusch, line 288: Same here. Shouldn't have to do that.

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    /// Accumulates the full body in-memory into a single buffer
+    /// up to `maxBytes` and returns it.
+    /// - Parameters:
+    ///   - maxBytes: The maximum number of bytes this method is allowed
+    ///     to accumulate in memory before it throws an error.
+    /// - Throws: `TooManyBytesError` if the the sequence contains more
+    ///   than `maxBytes`.
+    public func collect(upTo maxBytes: Int) async throws -> DataType {
```
FranzBusch, line 282: FWIW in async-altos we spelled this like that [AsyncSyncSequence](https://github.com/apple/swift-async-algorithms/blob/4a1fb99f0089a9d9db07859bcad55b4a77e3c3dd/Sources/AsyncAlgorithms/AsyncSyncSequence.swift#L17). Maybe we should just add a variant with a `upTo` in async-algos.

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+    public convenience init(
+        stream: AsyncStream<some StringProtocol>,
+        length: Body.Length
+    ) {
+        self.init(
+            sequence: .init(stream.map(\.asBodyChunk)),
+            length: length,
+            iterationBehavior: .single
+        )
+    }
```
FranzBusch, line 390: Same here as above.

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+    /// up to `maxBytes`, converts it to Foundation.Data, and returns it.
+    /// - Parameters:
+    ///   - maxBytes: The maximum number of bytes this method is allowed
+    ///     to accumulate in memory before it throws an error.
+    /// - Throws: `TooManyBytesError` if the the body contains more
+    ///   than `maxBytes`.
+    public func collectAsData(upTo maxBytes: Int) async throws -> Data {
+        let bytes: DataType = try await collect(upTo: maxBytes)
+        return Data(bytes)
+    }
```
FranzBusch, line 461: IMO this doesn't carry its weight for me since it is a one-liner for the user.

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+
+        private let produceNext: () async throws -> Element?
+
+        init<Iterator: AsyncIteratorProtocol>(
+            _ iterator: Iterator
+        ) where Iterator.Element == Element {
+            var iterator = iterator
+            self.produceNext = {
+                try await iterator.next()
+            }
```
FranzBusch, line 481: Why are we not storing the iterator here but a closure?

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+
+extension Body {
+
+    /// A type-erased async sequence that wraps input sequences.
+    private struct BodySequence: AsyncSequence {
+
+        typealias AsyncIterator = Body.Iterator
+        typealias Element = DataType
+
+        private let produceIterator: () -> AsyncIterator
```
FranzBusch, line 498: Same here

Sources/OpenAPIRuntime/Interface/Body.swift
```diff
@@ -0,0 +1,535 @@
...
+            }
+        }
+
+        func makeAsyncIterator() -> AsyncIterator {
+            produceIterator()
+        }
+    }
+
+    /// A wrapper for a sync sequence.
+    private struct WrappedSyncSequence<S: Sequence>: AsyncSequence
```
FranzBusch, line 512: FWIW we have this in async-algorithms.

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: > Why is this a class?

Because the body is inherently a reference type that cannot be freely copied.

> Also can we get rid of the @unchecked Sendable here and rather make ourselves a LockedValueBox abstraction that is conditionally Sendable when its contents are. This way we are making sure we really are Sendable.

Sure, that's an option that we can take as well, fortunately that won't affect the public API, so we can do that at any time.

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: Because API evolution? If so, I'd happily mark this as `@frozen` or give some other hint that we won't be expanding the enum.

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: Yes, for example any body created from a sync collection can be iterated multiple times. An async sequence for reading a file could also be legal to iterate multiple times. Basically, we want to provide the tools for those who want to support e.g. retryable file uploads.

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: Because API evolution? If so, I'd happily mark this as `@frozen` or give some other hint that we won't be expanding the enum.

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: We do, as this type will be owned by generated Equatable/Hashable types.

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: It's subtle, but we have them here because we know that async streams are _never_ safe to iterate multiple times (so we hardcode `.single` here, whereas a generic AsyncSequence might be, depends on the sequence and that's why those initializers force the user to specify either way.

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: That's a fair request, and something we also considered. However I find that the way that crash happens is pretty opaque and doesn't really help the user understand what went wrong, and how to fix it. We're just catching this scenario early and emitting a more descriptive error, which I think is generally a good approach? WDYT?

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: A common one: logging, a less common one: streaming encryption/unencryption/zipping/unzipping, etc in a middleware. That's why the result value still needs to be a body.

Outside of middlewares, you can just use the normal existing algorithms on AsyncSequence and get `A<B<...<MySequence>>>`, but in middlewares, the type of the sequence we provide to you, and that you provide back to us, has to be `Body`.

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: Agreed the wording needs work. The intention is: you specify the maximum size you're willing to buffer, and if the real body is larger, you get an error thrown (to protect your memory limits).

What terminology would you use here instead?

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: Fair, again here, it's meant to provide a better, more descriptive error that helps the adopter know what's wrong and how to fix it. Also, it throws an error here instead of crashing (as `makeAsyncIterator` isn't throwing, so we can't throw there), which again means that when the unlikely case happens (retry, redirect), the adopter's server/app isn't taken down, but they just get an error in their logs.

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: Sure - re depending on async-algos, we could do that once it has a 1.0, but not before.

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: Same answer - `iterationBehavior: .single` is why this initializer is helpful.

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: It was meant to mirror `collectAsString` in a way, and tries to help with systems where Data is the very common data type (iOS apps). Plus, it'll be called from generated code, from where we really try to only make simple calls, as opposed to embedding too much logic.

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: The intent here is type-erasure, is there a better way than to go through a closure?

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: https://github.com/apple/swift-openapi-runtime/pull/46#discussion_r1314603403

Sources/OpenAPIRuntime/Interface/Body.swift
czechboy0: https://github.com/apple/swift-openapi-runtime/pull/46#discussion_r1314597162

Sources/OpenAPIRuntime/Interface/Body.swift
dnadoba: Even if `AsyncIteratorProtocol` had primary associated type, we would prefer a closure as it will get specialised but an existential not, as of today.

Sources/OpenAPIRuntime/Interface/HTTPBody.swift
```diff
@@ -0,0 +1,535 @@
...
+        case .unknown:
+            validatedTransform = transform
+        }
+        return HTTPBody(
+            sequence: map(validatedTransform),
+            length: length,
+            iterationBehavior: iterationBehavior
+        )
+    }
+}
```
dnadoba, line 242: What's the use case for this method?

Sources/OpenAPIRuntime/Interface/HTTPBody.swift
```diff
@@ -0,0 +1,535 @@
...
+            buffer.append(contentsOf: chunk)
+        }
+        return buffer
+    }
+}
+
+// MARK: - String-based bodies
+
+extension StringProtocol {
+    fileprivate var asBodyChunk: HTTPBody.DataType {
```
dnadoba, line 314: `@inlinable` or otherwise quite slow as this is a generic over `StringProtocol`

dnadoba (commented):
`@inlinable` needs to be added to everything that is generic or it will be very slow

Sources/OpenAPIRuntime/Interface/HTTPBody.swift
czechboy0: https://github.com/apple/swift-openapi-runtime/pull/46#discussion_r1314593893

Sources/OpenAPIRuntime/Interface/HTTPBody.swift
czechboy0: This is an internal symbol, I thought the compiler doesn't need inlinable annotations for internal symbols, only public ones? (It should be inlining these even without the annotation I thought.)

czechboy0:
> `@inlinable` needs to be added to everything that is generic or it will be very slow

Everything that's generic _and public_, right?

dnadoba:
And internal functions if they are called from public generic functions and forward the generic type to internal functions.

Sources/OpenAPIRuntime/Interface/HTTPBody.swift
dnadoba: You call this in a public generic function (`HTTPBody.init(data: some StringProtocol, ...)`) which forwards the generic type to this function call. Therefore it needs to be @inlinable or it can't be specialised.

Sources/OpenAPIRuntime/Interface/HTTPBody.swift
dnadoba: All these use cases (streaming encryption/unencryption/zipping/unzipping) will not work as they will all change the chunk length. These are also stateful operations and the `@Sendable` requirement will make this needlessly hard. In addition they will usually buffer data and will need an explicit flush at the end which is not possible with this API. These use cases should just be modelled as `AsyncSequence` operators/transformers.

czechboy0: Yeah I can be convinced, I'll remove the API, it's not pulling its weight.

czechboy0:
@dnadoba @FranzBusch Let's move the conversation to the PR that has this plus the new currency types, as that's where I put more usage examples: https://github.com/apple/swift-openapi-runtime/pull/47/files#diff-e089a09ba06f85ed1377a6d7ec2a0c36a9f91a3d1b276134aa3156ec6484cede

PR #48, opened 2023-09-06
Fix optional query param parsing

czechboy0:
### Motivation

Fixes https://github.com/apple/swift-openapi-generator/issues/249.

### Modifications

Adds a `decodeIfPresent` method on `URIDecoder` now used by the optional query param getter.

### Result

This makes the query param getter return nil instead of throwing an error when it encounters a missing optional query param.

### Test Plan

Added new unit tests, hopefully these cover it now.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
@@ -134,7 +134,7 @@ extension Converter {
                         dateTranscoder: configuration.dateTranscoder
                     )
                 )
-                let value = try decoder.decode(
+                let value = try decoder.decodeIfPresent(
                     T.self,
                     forKey: name,
                     from: query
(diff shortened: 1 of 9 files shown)


czechboy0:

Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
```diff
@@ -134,7 +134,7 @@ extension Converter {
                         dateTranscoder: configuration.dateTranscoder
                     )
                 )
-                let value = try decoder.decode(
+                let value = try decoder.decodeIfPresent(
```
czechboy0, line 137: This is the core of the fix - notice we're in `getOptionalQueryItemAsURI`, but we were decoding unconditionally here.

tib (approved):
Looks good to me. Thank you for the fix.

czechboy0:

czechboy0:
@tib thanks for the review. Have you also had a chance to try this branch to verify the issue is fixed for the cases you tested?

czechboy0:

tib:
> @tib thanks for the review. Have you also had a chance to try this branch to verify the issue is fixed for the cases you tested?

Yes, I was able to verify it and the fix works just fine. If you could release this as a patch (0.2.1?) that'd be so nice. Thank you very much for the help. 🙏

glbrntt (approved):

simonjbeaumont (approved):

PR #55, opened 2023-09-26
Support base64-encoded data

rnro:
### Motivation

OpenAPI supports base64-encoded data but to this point OpenAPI Generator has not (https://github.com/apple/swift-openapi-generator/issues/11).

### Modifications

Introduce the `Base64EncodedData` codable type to allow users in the generator to describe byte types which must be en/de-coded.

### Result

Users will be able to describe base64-encoded data as `OpenAPIRuntime.Base64EncodedData` e.g.
```
public typealias MyData = OpenAPIRuntime.Base64EncodedData
```

### Test Plan

Added a round-trip encode/decode test `testEncodingDecodingRoundTrip_base64_success`

### DIFF
--- a/Sources/OpenAPIRuntime/Base/Base64EncodedData.swift
+++ b/Sources/OpenAPIRuntime/Base/Base64EncodedData.swift
@@ -1,0 +1,38 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+public struct Base64EncodedData: Sendable, Codable, Hashable {
+    var data: Foundation.Data // or [UInt8] or ArraySlice<UInt8>
+
+    public init(data: Foundation.Data) {
+        self.data = data
+    }
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        let base64EncodedString = try container.decode(String.self)
+        guard let data = Data(base64Encoded: base64EncodedString) else {
+            throw RuntimeError.invalidBase64String(base64EncodedString)
+        }
+        self.init(data: data)
+    }
+
+    public func encode(to encoder: any Encoder) throws {
+        var container = encoder.singleValueContainer()
+        let base64String = data.base64EncodedString()
+        try container.encode(base64String)
+    }
+}
--- a/Sources/OpenAPIRuntime/Base/Base64EncodedData.swift
+++ b/Sources/OpenAPIRuntime/Base/Base64EncodedData.swift
@@ -81,7 +81,6 @@ extension Base64EncodedData: Codable {
         self.init(data: ArraySlice(data))
     }
 
-
     /// Encodes this value into the given encoder.
     ///
     /// If the value fails to encode anything, `encoder` will encode an empty
--- a/Sources/OpenAPIRuntime/Base/Base64EncodedData.swift
+++ b/Sources/OpenAPIRuntime/Base/Base64EncodedData.swift
@@ -32,7 +32,7 @@ import Foundation
 /// let base64EncodedData = Base64EncodedData(from: decoder)
 ///```
 ///
-/// However more commonly the decoding initializer would be called by a decoder e.g.
+/// However more commonly the decoding initializer would be called by a decoder, for example:
 /// ```swift
 /// let encodedData: Data = ...
 /// let decoded = try JSONDecoder().decode(Base64EncodedData.self, from: encodedData)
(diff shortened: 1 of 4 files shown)


Tests/OpenAPIRuntimeTests/Base/Test_OpenAPIValue.swift
```diff
@@ -266,4 +266,13 @@ final class Test_OpenAPIValue: Test_Runtime {
         let nestedValue = try XCTUnwrap(nestedDict["nested"] as? Int)
         XCTAssertEqual(nestedValue, 2)
     }
+
+    func testEncodingDecodingRoundTrip_base64_success() throws {
+        print(String(data: testStructData.base64EncodedData(), encoding: .utf8)!.utf8)
+        let encodedData = Base64EncodedData(data: testStructData)
+        XCTAssertEqual(
+            try JSONDecoder().decode(Base64EncodedData.self, from: JSONEncoder().encode(encodedData)),
```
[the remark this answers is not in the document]
rnro: The reason I built it this way is that it was tricky to get hold of an `Encoder`/`Decoder` to inject into the `Base64EncodedData` type directly. 

I confirmed that the code as written calls through to the `init` and `encode` functions of `Base64EncodedData`, I presume delegated by the `JSONEncoder` when it encounters the type.

rnro:
> By users, do we mean the generator will emit this type alias when it encounters `format: base64` in the OpenAPI document?

Sorry, yes, the generator. I _think_ we will use `format: byte` though as that's mentioned in the [3.03 spec](https://spec.openapis.org/oas/v3.0.3)

czechboy0:

rnro:

Sources/OpenAPIRuntime/Base/Base64EncodedData.swift
```diff
@@ -0,0 +1,49 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+public struct Base64EncodedData: Sendable, Codable, Hashable {
```
czechboy0, line 17: Please add docs for all public symbols, ideally also with info how to use this type (create and consume).

Sources/OpenAPIRuntime/Base/Base64EncodedData.swift
```diff
@@ -0,0 +1,105 @@
...
+/// let bytes: ArraySlice<UInt8> = ...
+/// let base64EncodedData = Base64EncodedData(data: bytes)
+/// ```
+///
+/// To decode base64-encoded data it is possible to call the initializer directly, providing a decoder:
+/// ```swift
+/// let base64EncodedData = Base64EncodedData(from: decoder)
+///```
+///
+/// However more commonly the decoding initializer would be called by a decoder e.g.
```
czechboy0, line 35:
```suggestion
/// However more commonly the decoding initializer would be called by a decoder, for example:
```

Sources/OpenAPIRuntime/Base/Base64EncodedData.swift
```diff
@@ -0,0 +1,105 @@
...
+/// base64EncodedData.encode(to: encoder)
+/// ```
+///
+/// However more commonly it would be called by an encoder e.g.
+/// ```swift
+/// let bytes: ArraySlice<UInt8> = ...
+/// let encodedData = JSONEncoder().encode(encodedBytes)
+/// ```
+public struct Base64EncodedData: Sendable, Hashable {
+    /// data to be encoded
```
czechboy0, line 54:
```suggestion
    /// A container of the raw bytes.
```

Sources/OpenAPIRuntime/Base/Base64EncodedData.swift
```diff
@@ -0,0 +1,105 @@
...
+}
+
+extension Base64EncodedData: Codable {
+    /// Creates a new instance by decoding from the given decoder.
+    ///
+    /// This initializer throws an error if reading from the decoder fails, or
+    /// if the data read is corrupted or otherwise invalid.
+    ///
+    /// - Parameter decoder: The decoder to read data from.
+    public init(from decoder: any Decoder) throws {
```
czechboy0, line 71:
```suggestion
    public init(from decoder: any Decoder) throws {
```

We skip documenting conformances.

Sources/OpenAPIRuntime/Base/Base64EncodedData.swift
```diff
@@ -0,0 +1,105 @@
...
+    /// Encodes this value into the given encoder.
+    ///
+    /// If the value fails to encode anything, `encoder` will encode an empty
+    /// keyed container in its place.
+    ///
+    /// This function throws an error if any values are invalid for the given
+    /// encoder's format.
+    ///
+    /// - Parameter encoder: The encoder to write data to.
+    public func encode(to encoder: any Encoder) throws {
```
czechboy0, line 93:
```suggestion
    public func encode(to encoder: any Encoder) throws {
```

Same here.

Sources/OpenAPIRuntime/Errors/RuntimeError.swift
```diff
@@ -73,6 +74,8 @@ internal enum RuntimeError: Error, CustomStringConvertible, LocalizedError, Pret
             return "Invalid expected content type: '\(string)'"
         case .invalidHeaderFieldName(let name):
             return "Invalid header field name: '\(name)'"
+        case .invalidBase64String(let string):
+            return "Invalid base64-encoded string: '\(string)'"
```
czechboy0, line 78: Let's cap how many bytes we'd actually print.
```suggestion
            return "Invalid base64-encoded string (first 128 bytes): '\(string.prefix(128))'"
```

czechboy0 (changes requested):
A few minor requests, otherwise looks good! Could you also open a PR of the generator changes, so that we can review both before we start landing them? That's our usual workflow. That means that locally you need to develop the two branches together.

rnro:
The accompanying generator change is https://github.com/apple/swift-openapi-generator/pull/326

czechboy0:

czechboy0 (approved):
Lgtm, just rerunning the integration test which should pass now that 0.3.0 has been released.

simonjbeaumont (approved):
Nice—thanks @rnro !

PR #58, opened 2023-10-11
[Runtime] Support recursive types

czechboy0:
### Motivation

Runtime changes for https://github.com/apple/swift-openapi-generator/issues/70.

### Modifications

Introduce a new wrapper type `CopyOnWriteBox` as a reference type with copy-on-write semantics, to be used by the generated code to hide a ref type inside a struct that needs boxing. (Details on the logic will be in the generator PR description.)

### Result

A new type available for the generated code to use for breaking reference cycles.

### Test Plan

Added unit tests for the type.

### DIFF
--- a/Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
+++ b/Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
@@ -12,8 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-/// A wrapper reference type for a value that should be copied only when
-/// referenced by multiple owners and is being modified.
+/// A reference type that wraps a value and enforces copy-on-write semantics.
 ///
 /// It also enables recursive types by introducing a "box" into the cycle, which
 /// allows the owning type to have a finite size.
--- a/Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
+++ b/Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
@@ -55,12 +55,31 @@ public final class CopyOnWriteBox<Wrapped> {
 }
 
 extension CopyOnWriteBox: Encodable where Wrapped: Encodable {
+
+    /// Encodes this value into the given encoder.
+    ///
+    /// If the value fails to encode anything, `encoder` will encode an empty
+    /// keyed container in its place.
+    ///
+    /// This function throws an error if any values are invalid for the given
+    /// encoder's format.
+    ///
+    /// - Parameter encoder: The encoder to write data to.
+    /// - Throws: On an encoding error.
     public func encode(to encoder: any Encoder) throws {
         try value.encode(to: encoder)
     }
 }
 
 extension CopyOnWriteBox: Decodable where Wrapped: Decodable {
+
+    /// Creates a new instance by decoding from the given decoder.
+    ///
+    /// This initializer throws an error if reading from the decoder fails, or
+    /// if the data read is corrupted or otherwise invalid.
+    ///
+    /// - Parameter decoder: The decoder to read data from.
+    /// - Throws: On a decoding error.
     public convenience init(from decoder: any Decoder) throws {
         let value = try Wrapped(from: decoder)
         self.init(value: value)
@@ -68,6 +87,16 @@ extension CopyOnWriteBox: Decodable where Wrapped: Decodable {
 }
 
 extension CopyOnWriteBox: Equatable where Wrapped: Equatable {
+
+    /// Returns a Boolean value indicating whether two values are equal.
+    ///
+    /// Equality is the inverse of inequality. For any values `a` and `b`,
+    /// `a == b` implies that `a != b` is `false`.
+    ///
+    /// - Parameters:
+    ///   - lhs: A value to compare.
+    ///   - rhs: Another value to compare.
+    /// - Returns: A Boolean value indicating whether the values are equal.
     public static func == (
         lhs: CopyOnWriteBox<Wrapped>,
         rhs: CopyOnWriteBox<Wrapped>
@@ -77,18 +106,82 @@ extension CopyOnWriteBox: Equatable where Wrapped: Equatable {
 }
 
 extension CopyOnWriteBox: Hashable where Wrapped: Hashable {
+
+    /// Hashes the essential components of this value by feeding them into the
+    /// given hasher.
+    ///
+    /// Implement this method to conform to the `Hashable` protocol. The
+    /// components used for hashing must be the same as the components compared
+    /// in your type's `==` operator implementation. Call `hasher.combine(_:)`
+    /// with each of these components.
+    ///
+    /// - Important: In your implementation of `hash(into:)`,
+    ///   don't call `finalize()` on the `hasher` instance provided,
+    ///   or replace it with a different instance.
+    ///   Doing so may become a compile-time error in the future.
+    ///
+    /// - Parameter hasher: The hasher to use when combining the components
+    ///   of this instance.
     public func hash(into hasher: inout Hasher) {
         hasher.combine(value)
     }
 }
 
 extension CopyOnWriteBox: CustomStringConvertible where Wrapped: CustomStringConvertible {
+
+    /// A textual representation of this instance.
+    ///
+    /// Calling this property directly is discouraged. Instead, convert an
+    /// instance of any type to a string by using the `String(describing:)`
+    /// initializer. This initializer works with any type, and uses the custom
+    /// `description` property for types that conform to
+    /// `CustomStringConvertible`:
+    ///
+    ///     struct Point: CustomStringConvertible {
+    ///         let x: Int, y: Int
+    ///
+    ///         var description: String {
+    ///             return "(\(x), \(y))"
+    ///         }
+    ///     }
+    ///
+    ///     let p = Point(x: 21, y: 30)
+    ///     let s = String(describing: p)
+    ///     print(s)
+    ///     // Prints "(21, 30)"
+    ///
+    /// The conversion of `p` to a string in the assignment to `s` uses the
+    /// `Point` type's `description` property.
     public var description: String {
         value.description
     }
 }
 
 extension CopyOnWriteBox: CustomDebugStringConvertible where Wrapped: CustomDebugStringConvertible {
+
+    /// A textual representation of this instance, suitable for debugging.
+    ///
+    /// Calling this property directly is discouraged. Instead, convert an
+    /// instance of any type to a string by using the `String(reflecting:)`
+    /// initializer. This initializer works with any type, and uses the custom
+    /// `debugDescription` property for types that conform to
+    /// `CustomDebugStringConvertible`:
+    ///
+    ///     struct Point: CustomDebugStringConvertible {
+    ///         let x: Int, y: Int
+    ///
+    ///         var debugDescription: String {
+    ///             return "(\(x), \(y))"
+    ///         }
+    ///     }
+    ///
+    ///     let p = Point(x: 21, y: 30)
+    ///     let s = String(reflecting: p)
+    ///     print(s)
+    ///     // Prints "(21, 30)"
+    ///
+    /// The conversion of `p` to a string in the assignment to `s` uses the
+    /// `Point` type's `debugDescription` property.
     public var debugDescription: String {
         value.debugDescription
     }
--- a/Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
+++ b/Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
@@ -20,28 +20,34 @@
 public struct CopyOnWriteBox<Wrapped> {
 
     /// The reference type storage for the box.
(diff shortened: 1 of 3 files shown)


Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
```diff
@@ -0,0 +1,190 @@
...
+    /// Returns a Boolean value indicating whether two values are equal.
+    ///
+    /// Equality is the inverse of inequality. For any values `a` and `b`,
+    /// `a == b` implies that `a != b` is `false`.
+    ///
+    /// - Parameters:
+    ///   - lhs: A value to compare.
+    ///   - rhs: Another value to compare.
+    /// - Returns: A Boolean value indicating whether the values are equal.
+    public static func == (
```
dnadoba, line 100: Do we really want to copy & paste the documentation for these? It seems DocC doesn't currently auto inherit the documentation but this is more of a tooling problem.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
```diff
@@ -0,0 +1,190 @@
...
+    ) {
+        let resolvedBox: CopyOnWriteBox<Wrapped>
+        if isKnownUniquelyReferenced(&box) {
+            resolvedBox = box
+        } else {
+            resolvedBox = Self(value: box.value)
+            box = resolvedBox
+        }
+        modify(&resolvedBox.value)
+    }
```
dnadoba, line 54: Why is this a static method instead of an instance method?

dnadoba: Have you experimented with exposing the underlying value as computed property and enforcing uniqueness in a `_modify` block? I think this should work as well and would provide a nice interface.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
czechboy0: This is to appease swift-format, we're strict about all public symbols having docs. I hope there'll be a feature in the future that allows omitting this and inherit from the protocol, but until then, this is fine.

dnadoba: hm... not great as this more than 50% of the whole file and will get outdated eventually 😅 but yeah, not a blocker.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
czechboy0: So the reason for the static func is because we need to write back into the box, and exclusivity rules prohibit the "obvious" solution of passing a ref to self and also capturing it in the closure.

I haven't tried `_modify` to avoid underscored APIs so far. Note that this is never used directly, only generated code calls this, so clarity is a bit more important here than ergonomics.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
czechboy0: Hopefully the eventual support could actually _warn_ you if you've documented a symbol documented outside your type. That way it'd tell you all the things you can remove.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
dnadoba: Why aren't you passing `Value` `inout` instead of `Self`?

dnadoba: Oh I see, you need to create a new box on copy. That makes sense.

dnadoba: So you can make it a instance method it if you wrap the class in a struct:
```swift
public struct CopyOnWriteBox<Wrapped> {
    private final class Storage {
        var value: Wrapped
        
        init(value: Wrapped) {
            self.value = value
        }
    }
    
    private var storage: Storage
    
    public init(value: Wrapped) {
        self.storage = .init(value: value)
    }
    
    public func read() -> Wrapped {
        storage.value
    }
    
    public mutating func write(
        using modify: (inout Wrapped) -> Void
    ) {
        let resolvedStorage: Storage
        if isKnownUniquelyReferenced(&storage) {
            resolvedStorage = storage
        } else {
            resolvedStorage = Storage(value: storage.value)
            self.storage = resolvedStorage
        }
        modify(&resolvedStorage.value)
    }
}
```

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
dnadoba: I disagree. Sometimes you want and need to overwrite the documentation. A good example is the `subscript(position: Index) -> Element` form `Collection`. You usually want to state the time complexity of the subscript as it defaults to `O(n)` if it is better (or worse).

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
dnadoba: Nit: You also usually want to make these methods rethrows

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
czechboy0: I'd probably instead put that info into the docs of the type, not the method coming from e.g. a protocol as I'd consider performance characteristics part of the protocol's API, and any divergence should be known at value creation time, because generics/existentials allow you to use different types the same way. So when I'm calling this method, I might not know which type I have - the protocol definition is all I have when deciding how to use the value.

But this is pretty nuanced and sure, an option to warn on documented overrides would surely be optional, not required 🙂

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
czechboy0: We do have a wrapper struct, that's generated. This code isn't used manually, it's SPI for the generated code, so we should consider it in the context of its call site. If you're interested in how it's used, check out the mirror PR on the generator for recursive type support.

czechboy0:
Will move to `get` + `_modify` tomorrow, so taking the PR back to draft.

czechboy0:
@dnadoba Applied your feedback, can you take another look?

dnadoba (changes requested):
Everything needs to be `@inlineable`, including the protocol conformances.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
```diff
@@ -0,0 +1,191 @@
...
+    private var storage: Storage
+
+    /// Creates a new box.
+    /// - Parameter value: The value to store in the box.
+    public init(value: Wrapped) {
+        self.storage = .init(value: value)
+    }
+
+    /// The stored value whose accessors enforce copy-on-write semantics.
+    public var value: Wrapped {
```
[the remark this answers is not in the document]
czechboy0: Added: https://github.com/apple/swift-openapi-runtime/pull/58/commits/cf4ee660bb914e014058fc1b9ca25b53d45a75e1

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
czechboy0: Resolving as this was completely changed.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
```diff
@@ -0,0 +1,197 @@
...
+        /// The stored value.
+        @usableFromInline
+        var value: Wrapped
+
+        /// Creates a new storage with the provided initial value.
+        /// - Parameter value: The initial value to store in the box.
+        @usableFromInline
+        init(value: Wrapped) {
+            self.value = value
+        }
```
dnadoba, line 35: `Storage` is implicitly generic because its outer type is generic so this should be `@inlinable` as well.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
```diff
@@ -0,0 +1,197 @@
...
+    /// with each of these components.
+    ///
+    /// - Important: In your implementation of `hash(into:)`,
+    ///   don't call `finalize()` on the `hasher` instance provided,
+    ///   or replace it with a different instance.
+    ///   Doing so may become a compile-time error in the future.
+    ///
+    /// - Parameter hasher: The hasher to use when combining the components
+    ///   of this instance.
+    public func hash(into hasher: inout Hasher) {
```
dnadoba, line 132: `@inlinable` is missing.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
dnadoba: `@inlinable` is missing.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
```diff
@@ -0,0 +1,197 @@
...
+extension CopyOnWriteBox: Decodable where Wrapped: Decodable {
+
+    /// Creates a new instance by decoding from the given decoder.
+    ///
+    /// This initializer throws an error if reading from the decoder fails, or
+    /// if the data read is corrupted or otherwise invalid.
+    ///
+    /// - Parameter decoder: The decoder to read data from.
+    /// - Throws: On a decoding error.
+    public init(from decoder: any Decoder) throws {
```
dnadoba, line 90: `@inlinable` missing.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
```diff
@@ -0,0 +1,197 @@
...
+    ///
+    /// If the value fails to encode anything, `encoder` will encode an empty
+    /// keyed container in its place.
+    ///
+    /// This function throws an error if any values are invalid for the given
+    /// encoder's format.
+    ///
+    /// - Parameter encoder: The encoder to write data to.
+    /// - Throws: On an encoding error.
+    public func encode(to encoder: any Encoder) throws {
```
dnadoba, line 76: `@inlinable` missing.

dnadoba (changes requested):
Some `@inlinable` are missing but otherwise it looks good.
Note that the implementations of the` Custom[Debug]StringConvertible` protocols could be made `@inlinable` as well but they are usually called through existential anyway so I'm fine without adding them.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
czechboy0: https://github.com/apple/swift-openapi-runtime/pull/58/commits/9f1e5cdd83c2e434c8c33b8d6dbf9895f19a7f70

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
czechboy0: https://github.com/apple/swift-openapi-runtime/pull/58/commits/9f1e5cdd83c2e434c8c33b8d6dbf9895f19a7f70

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
czechboy0: https://github.com/apple/swift-openapi-runtime/pull/58/commits/9f1e5cdd83c2e434c8c33b8d6dbf9895f19a7f70

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
czechboy0: https://github.com/apple/swift-openapi-runtime/pull/58/commits/9f1e5cdd83c2e434c8c33b8d6dbf9895f19a7f70

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
```diff
@@ -0,0 +1,191 @@
...
+    public var value: Wrapped {
+        get {
+            storage.value
+        }
+        _modify {
+            if !isKnownUniquelyReferenced(&storage) {
+                storage = Storage(value: storage.value)
+            }
+            yield &storage.value
+        }
```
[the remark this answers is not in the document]
czechboy0: Talked offline, not needed it seems.

Sources/OpenAPIRuntime/Base/CopyOnWriteBox.swift
czechboy0: https://github.com/apple/swift-openapi-runtime/pull/58/commits/9f1e5cdd83c2e434c8c33b8d6dbf9895f19a7f70

dnadoba (approved):

PR #63, opened 2023-10-25
Improved enriched error propagation from the transport and middlewares

czechboy0:
### Motivation

Fixes https://github.com/apple/swift-openapi-generator/issues/302 and https://github.com/apple/swift-openapi-generator/issues/17.

The issue was that we hid away errors thrown in transports and middlewares, and the adopter would get `ClientError` where the `underlyingError` wasn't the error thrown by the underlying transport/middleware, but instead a private wrapper of ours.

### Modifications

Make sure `{Client,Server}Error.underlyingError` contains the error thrown from the underlying transport/middleware when that was the cause of the error, otherwise keep `RuntimeError` there as was the behavior until now.

Also added a `causeDescription` property on both public error types to allow communicating the context for the underlying error.

Also made sure middleware errors are now wrapped in Client/ServerError, they weren't before so didn't contain the context necessary to debug issues well.

### Result

Adopters can now extract the errors thrown e.g. by URLSession from our public error types using the `underlyingError` property and understand the context of where it was thrown by checking the user-facing `causeDescription`.

Also, adopters now get enriched errors thrown from middlewares.

### Test Plan

Wrote unit tests for both UniversalClient and UniversalServer, inevitably found some minor bugs there as well, fixed them all, plus the unit tests now verify the behavior new in this PR.

### DIFF
--- a/Sources/OpenAPIRuntime/Errors/ClientError.swift
+++ b/Sources/OpenAPIRuntime/Errors/ClientError.swift
@@ -64,6 +64,10 @@ public struct ClientError: Error {
     /// Will be nil if the error resulted before the response was received.
     public var responseBody: HTTPBody?
 
+    /// A user-facing description of what caused the underlying error
+    /// to be thrown.
+    public var causeDescription: String
+
     /// The underlying error that caused the operation to fail.
     public var underlyingError: any Error
 
@@ -76,6 +80,8 @@ public struct ClientError: Error {
     ///   - baseURL: The base URL for HTTP requests.
     ///   - response: The HTTP response received during the operation.
     ///   - responseBody: The HTTP response body received during the operation.
+    ///   - causeDescription: A user-facing description of what caused
+    ///     the underlying error to be thrown.
     ///   - underlyingError: The underlying error that caused the operation
     ///     to fail.
     public init(
@@ -86,6 +92,7 @@ public struct ClientError: Error {
         baseURL: URL? = nil,
         response: HTTPResponse? = nil,
         responseBody: HTTPBody? = nil,
+        causeDescription: String,
         underlyingError: any Error
     ) {
         self.operationID = operationID
@@ -95,6 +102,7 @@ public struct ClientError: Error {
         self.baseURL = baseURL
         self.response = response
         self.responseBody = responseBody
+        self.causeDescription = causeDescription
         self.underlyingError = underlyingError
     }
 
@@ -115,7 +123,7 @@ extension ClientError: CustomStringConvertible {
     ///
     /// - Returns: A string describing the client error and its associated details.
     public var description: String {
-        "Client error - operationID: \(operationID), operationInput: \(String(describing: operationInput)), request: \(request?.prettyDescription ?? "<nil>"), requestBody: \(requestBody?.prettyDescription ?? "<nil>"), baseURL: \(baseURL?.absoluteString ?? "<nil>"), response: \(response?.prettyDescription ?? "<nil>"), responseBody: \(responseBody?.prettyDescription ?? "<nil>") , underlying error: \(underlyingErrorDescription)"
+        "Client error - cause description: '\(causeDescription)', underlying error: \(underlyingErrorDescription), operationID: \(operationID), operationInput: \(String(describing: operationInput)), request: \(request?.prettyDescription ?? "<nil>"), requestBody: \(requestBody?.prettyDescription ?? "<nil>"), baseURL: \(baseURL?.absoluteString ?? "<nil>"), response: \(response?.prettyDescription ?? "<nil>"), responseBody: \(responseBody?.prettyDescription ?? "<nil>")"
     }
 }
 
--- a/Sources/OpenAPIRuntime/Interface/UniversalClient.swift
+++ b/Sources/OpenAPIRuntime/Interface/UniversalClient.swift
@@ -90,19 +90,20 @@ import Foundation
         serializer: @Sendable (OperationInput) throws -> (HTTPRequest, HTTPBody?),
         deserializer: @Sendable (HTTPResponse, HTTPBody?) async throws -> OperationOutput
     ) async throws -> OperationOutput where OperationInput: Sendable, OperationOutput: Sendable {
-        @Sendable
-        func wrappingErrors<R>(
+        @Sendable func wrappingErrors<R>(
             work: () async throws -> R,
             mapError: (any Error) -> any Error
         ) async throws -> R {
             do {
                 return try await work()
+            } catch let error as ClientError {
+                throw error
             } catch {
                 throw mapError(error)
             }
         }
         let baseURL = serverURL
-        func makeError(
+        @Sendable func makeError(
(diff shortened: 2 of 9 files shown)


czechboy0:
cc @groue @MahdiBM

MahdiBM (commented):
It appears to get the full description of the error we'll need to do something like `String(describing: error)` instead of just a string interpolation. 
That seems to be a trend as PostgresNIO already does that to make sure not to leak database info, but I was hoping for the latter in this library since i'm not sure if any secret info will be leaked by just showing a connection error to users mistakenly when you when string-interpolating the error.

Overall I'm fine with it as we've already moved to using `String(reflecting:)` for errors in most places, but I'm still not sure if it's the correct way for this library.

czechboy0:
Hmm I'm not sure I understand your concern. Can you elaborate on what you're trying to do, how the library makes that difficult, and what you'd like to see instead?

This PR is about allowing you to programmatically get the unmodified underlyingError. Is that not what you needed?

MahdiBM:
Sort of, but not exactly.

Now that i'm testing it, it actually doesn't solve my problem for the most par, at a;;.

Given this code:
```swift
let error: any Error = RuntimeError.transportFailed(
    DecodingError.dataCorrupted(.init(codingPath: [], debugDescription: "description!"))
)

print(error) // Transport threw an error.
print(String(describing: error)) // Transport threw an error.
logger.log("Got error", metadata: ["error": "\(error)"]) // Transport threw an error.
logger.log("Got error", metadata: ["error": "\(String(reflecting: error))"]) // Transport threw an error.
```
You see? It's not very debuggable. This is exactly my use-case.
One _could_ catch the error and print it, but IMO it's not right to expect am explicit catch at all times just to be able to debug.

Solution 1:
Add `CustomDebugStringConvertible` to the error types, and let it include everything.
`String(reflecting: error)` will use that, so result could look like:
```swift
print(error) // Transport threw an error.
print(String(describing: error)) // Transport threw an error: THE UNDERLYING ERR
logger.log("Got error", metadata: ["error": "\(error)"]) // Transport threw an error.
logger.log("Got error", metadata: ["error": "\(String(reflecting: error))"]) // Transport threw an error: THE UNDERLYING ERR
```

Solution 2:
Make the `CustomStringConvertible` contain the full error, in which case all the above print/logs will contain the full error.

MahdiBM:
Another note. I notice the edited types still conform to `LocalizedError` but don't provide the whole point of the conformance which is the `errorDescription` variable, unlike other error types.
Something should probably change. Either removal of the conformance, or providing the `errorDescription`. I think the latter would be what you want considering other Error types do that.

czechboy0:
RuntimeError is an internal type, so what you showed above is never what you'll get. You'll get a ClientError, where the causeDescription is "Transport threw an error." and the underlyingError is the actual error threw by the transport. When you print the ClientError, it should contain a summary of everything it includes, but you can always grab the underlyingError yourself and print or cast it to a concrete type and inspect if further.

Can you check out this branch and show, by using a generated Client and a mock ClientTransport that throws a custom error, how any part of the error is missing when you print it?

czechboy0:
Also note that because ClientError and ServerError are public types, you can create a completely custom way to print them, all their properties that store what's printed by `description` is public.

MahdiBM:
> RuntimeError is an internal type, .... You'll get a ClientError, 

That's actually great, sorry that i failed to remember/notice that. If what users always get is a `ClientError`, then this PR absolutely solves the problem I was having, considering you've added the `underlyingError` to its description.

czechboy0:
> > RuntimeError is an internal type ... You'll get a ClientError
> 
> That's actually great, sorry that i failed to remember/notice that. If what users _always_ get is a `ClientError`, then this PR absolutely solves the problem I was having, considering you've added the `underlyingError` to `ClientError`'s description.

Cool - yeah and this PR also fixes https://github.com/apple/swift-openapi-generator/issues/17 as previously, you wouldn't get a ClientError when a middleware threw, but the raw underlying error. This PR makes it all consistent, the generated client should always throw a ClientError, and the underlyingError either has the error thrown by the middleware/transport, or it has the internal runtime error which is only useful for logging. But in general, for logging, I recommend logging the full ClientError as its description contains both the context of where the error was thrown, plus the error itself. We only expect users who need to programmatically inspect the underlyingError to actually access that property.

czechboy0:
The API breakage CI pipeline is red but it's a bug in diagnose-api-breaking-changes, I believe, and filed as rdar://117520415. So we can consider the CI green.

Sources/OpenAPIRuntime/Interface/UniversalClient.swift
```diff
@@ -90,76 +90,116 @@ import Foundation
         serializer: @Sendable (OperationInput) throws -> (HTTPRequest, HTTPBody?),
         deserializer: @Sendable (HTTPResponse, HTTPBody?) async throws -> OperationOutput
     ) async throws -> OperationOutput where OperationInput: Sendable, OperationOutput: Sendable {
-        @Sendable
-        func wrappingErrors<R>(
+        @Sendable func wrappingErrors<R>(
```
glbrntt, line 93: Is this a result of changing formatting options or just a manual fix up?

Sources/OpenAPIRuntime/Errors/ClientError.swift
```diff
@@ -64,6 +64,10 @@ public struct ClientError: Error {
     /// Will be nil if the error resulted before the response was received.
     public var responseBody: HTTPBody?
 
+    /// A user-facing description of what caused the underlying error
+    /// to be thrown.
+    public var causeDescription: String
```
glbrntt, line 69: Why use a `causeDescription` over wrapping the actual underlying cause in another error which represents the cause description? 

For example, with the middleware case, I can imagine something like:

```swift
struct MiddlewareError: Error {
  var middlewareType: String
  var underlyingError: any Error

  var description: String { 
    "Middleware of type '(middlewareType)' threw ..."
  }
}
```

czechboy0: We'd have to introduce several new public types then, one for Transport, another for Middleware, and so on. But the use case doesn't require it. Folks just want to get the underlying root error to handle programmatically, and they want to be able to log the cause (but not handle it programmatically).

For code internal to a package I'd probably do what you suggest, but here since we're crossing an API stable boundary, we don't want to expand the surface unless it's really justified. (For example, we don't want folks programmatically handling the `middlewareType` for the same reason we made RuntimeError internal and not public - as it's inherently unstable, so it should only be consumed in logs).

Sources/OpenAPIRuntime/Interface/UniversalClient.swift
czechboy0: Manual, I believe, but soon when respectExistingLineBreaks is turned off in the runtime library (as it was recently in the generator), it won't matter, as it'll enforce one style regardless of how it was originally written. And I think swift-format will put it on a single line.

glbrntt (approved):

PR #64, opened 2023-10-26
[Runtime] Generate server variables

czechboy0:
### Motivation

The runtime side of https://github.com/apple/swift-openapi-generator/issues/24.

### Modifications

Added an SPI `ServerVariable` type and a variant of the `URL.init(validatingOpenAPIServerURL:variables:)` method, which takes the template and variables and returns a fully formed concrete URL.

### Result

Unblocked the generator half of supporting server variables.

### Test Plan

Added unit tests.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/ServerVariable.swift
+++ b/Sources/OpenAPIRuntime/Conversion/ServerVariable.swift
@@ -1,0 +1,77 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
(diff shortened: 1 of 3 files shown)


Sources/OpenAPIRuntime/Conversion/ServerVariable.swift
```diff
@@ -0,0 +1,77 @@
...
+        guard let url = Self(string: urlString) else {
+            throw RuntimeError.invalidServerURL(urlString)
+        }
+        self = url
+    }
+}
+
+/// A variable of a server URL template in the OpenAPI document.
+@_spi(Generated)
+public struct ServerVariable {
```
glbrntt, line 54: `Sendable` and `Hashable`?

glbrntt (approved):

Sources/OpenAPIRuntime/Conversion/ServerVariable.swift
czechboy0: Good point, added.

PR #65, opened 2023-10-27
[Runtime] Improved content type matching

czechboy0:
### Motivation

The runtime changes for https://github.com/apple/swift-openapi-generator/issues/315.

### Modifications

- Introduces a new SPI method `Converter.bestContentType` that takes a received content type value and from a provided list of other content types, picks the most appropriate one. This actually follows the specification now, by going from most specific (including parameter matching) to least specific (most wildcard-y).
- Deprecates the previously used methods `Converter.makeUnexpectedContentTypeError` and `Converter.isMatchingContentType`.

### Result

SPI methods that the generated code can use to correctly match content types.

### Test Plan

Added unit tests.

### DIFF
--- a/Sources/OpenAPIRuntime/Base/OpenAPIMIMEType.swift
+++ b/Sources/OpenAPIRuntime/Base/OpenAPIMIMEType.swift
@@ -279,7 +279,7 @@ extension OpenAPIMIMEType {
 
             // According to RFC 2045: https://www.rfc-editor.org/rfc/rfc2045#section-5.1
             // "Type, subtype, and parameter names are case-insensitive."
-            // Inferred: Parameter values are be case-sensitive.
+            // Inferred: Parameter values are case-sensitive.
 
             let receivedNormalizedParameters = Dictionary(
                 uniqueKeysWithValues: receivedParameters.map { ($0.key.lowercased(), $0.value) }
--- a/Sources/OpenAPIRuntime/Base/OpenAPIMIMEType.swift
+++ b/Sources/OpenAPIRuntime/Base/OpenAPIMIMEType.swift
@@ -187,3 +187,115 @@ extension OpenAPIMIMEType: LosslessStringConvertible {
             .joined(separator: "; ")
     }
 }
+
+// MARK: - Internals
+
+extension OpenAPIMIMEType {
+
+    /// The result of a match evaluation between two MIME types.
+    enum Match: Hashable {
+
+        /// The reason why two types are incompatible.
+        enum IncompatibilityReason: Hashable {
+
+            /// The types don't match.
+            case type
+
+            /// The subtypes don't match.
+            case subtype
+
+            /// The parameter of the provided name is missing or doesn't match.
+            case parameter(name: String)
+        }
+
+        /// The types are incompatible for the provided reason.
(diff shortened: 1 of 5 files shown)


Sources/OpenAPIRuntime/Base/OpenAPIMIMEType.swift
```diff
@@ -187,3 +187,115 @@ extension OpenAPIMIMEType: LosslessStringConvertible {
...
+            //   2. If an option parameter is not received, this is an
+            //      incompatible content type match.
+            // This means we can just iterate over option parameters and
+            // check them against the received parameters, but we can
+            // ignore any received parameters that didn't appear in the
+            // option parameters.
+
+            // According to RFC 2045: https://www.rfc-editor.org/rfc/rfc2045#section-5.1
+            // "Type, subtype, and parameter names are case-insensitive."
+            // Inferred: Parameter values are be case-sensitive.
```
gjcairo, line 282:
```suggestion
            // Inferred: Parameter values are case-sensitive.
```

Sources/OpenAPIRuntime/Conversion/Converter+Common.swift
```diff
@@ -29,45 +29,52 @@ extension Converter {
...
+                receivedType: receivedType,
+                receivedSubtype: receivedSubtype,
+                receivedParameters: received.parameters,
+                against: parsedOption
+            )
+            return (contentType: stringOption, match: match)
         }
-        guard let expectedContentType = OpenAPIMIMEType(expectedRaw) else {
-            throw RuntimeError.invalidExpectedContentType(expectedRaw)
+        let sortedOptions = evaluatedOptions.sorted { a, b in
```
gjcairo, line 69: This may be a needless optimisation, but we could use `Array/max(by:)` ([docs](https://developer.apple.com/documentation/swift/array/max(by:))) instead of sorting, since we're only getting the first option anyways.

Tests/OpenAPIRuntimeTests/Conversion/Test_Converter+Common.swift
```diff
@@ -54,6 +55,153 @@ final class Test_CommonConverterExtensions: Test_Runtime {
...
+            ],
+            expected: "application/json"
+        )
+        XCTAssertThrowsError(
+            try testCase(
+                received: "application/json",
+                options: [
+                    "whoops"
+                ],
+                expected: "application/json"
```
gjcairo, line 103: Small nit but maybe pass an empty string/nil when we expect errors to be thrown instead, for clarity?

gjcairo (approved):
Looks great overall, just left a few comments with some pretty small nits.

Sources/OpenAPIRuntime/Conversion/Converter+Common.swift
czechboy0: Good call, updated.

Tests/OpenAPIRuntimeTests/Conversion/Test_Converter+Common.swift
czechboy0: Good idea, updated.

PR #66, opened 2023-10-27
[Runtime] Include partial errors in oneOf/anyOf decoding errors

czechboy0:
### Motivation

The runtime changes to address https://github.com/apple/swift-openapi-generator/issues/275.

This makes debugging of decoding of oneOf/anyOf much easier, as the individual errors aren't dropped on the floor anymore.

### Modifications

Added SPI that allows the generated code to collect and report partial errors when a oneOf/anyOf fails to decode (that includes trying multiple subschemas, which themselves emit errors when they're not the right match).

### Result

Easier debugging of oneOf/anyOf decoding issues.

### Test Plan

Tested manually as part of the generator changes, we don't generally test exact error strings.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/CodableExtensions.swift
+++ b/Sources/OpenAPIRuntime/Conversion/CodableExtensions.swift
@@ -189,7 +189,7 @@ extension Encoder {
 }
 
 /// A freeform String coding key for decoding undocumented values.
-private struct StringKey: CodingKey, Hashable, Comparable {
+struct StringKey: CodingKey, Hashable, Comparable {
 
     var stringValue: String
     var intValue: Int? {
--- a/Sources/OpenAPIRuntime/Conversion/CodableExtensions.swift
+++ b/Sources/OpenAPIRuntime/Conversion/CodableExtensions.swift
@@ -189,7 +189,7 @@ extension Encoder {
 }
 
 /// A freeform String coding key for decoding undocumented values.
-struct StringKey: CodingKey, Hashable, Comparable {
+private struct StringKey: CodingKey, Hashable, Comparable {
 
     var stringValue: String
     var intValue: Int? {
--- a/Sources/OpenAPIRuntime/Conversion/ErrorExtensions.swift
+++ b/Sources/OpenAPIRuntime/Conversion/ErrorExtensions.swift
@@ -22,16 +22,19 @@ extension DecodingError {
     ///   occurred.
     ///   - codingPath: The coding path to the decoder that attempted to decode
     ///   the type.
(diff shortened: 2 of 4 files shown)


Sources/OpenAPIRuntime/Conversion/ErrorExtensions.swift
```diff
@@ -70,17 +98,49 @@ extension DecodingError {
...
+                codingPath: codingPath,
+                errors: errors
             )
         }
     }
 }
+
+/// A wrapper of multiple errors, for example collected during a parallelized
+/// operation from the individual subtasks.
+struct MultiError: Swift.Error, LocalizedError, CustomStringConvertible {
```
simonjbeaumont, line 122: Should this be `@_spi`?

Sources/OpenAPIRuntime/Conversion/CodableExtensions.swift
```diff
@@ -189,7 +189,7 @@ extension Encoder {
 }
 
 /// A freeform String coding key for decoding undocumented values.
-private struct StringKey: CodingKey, Hashable, Comparable {
+struct StringKey: CodingKey, Hashable, Comparable {
```
simonjbeaumont, line 192: Should this be `@_spi`?

Sources/OpenAPIRuntime/Conversion/ErrorExtensions.swift
czechboy0: It's internal, so no need.

Sources/OpenAPIRuntime/Conversion/CodableExtensions.swift
czechboy0: Same as above, internal.

czechboy0: Hmm actually this isn't needed anymore, let me revert.

czechboy0: https://github.com/apple/swift-openapi-runtime/pull/66/commits/5fa5ab097dc875756a05ae09437a07785decadb8

simonjbeaumont (approved):

czechboy0:
API breakage is red with:

> 1 breaking change detected in OpenAPIRuntime:
>  💔 API breakage: func DecodingError.verifyAtLeastOneSchemaIsNotNil(_:type:codingPath:) is now with @_spi
> ** ERROR: ❌ Breaking API changes detected.

This method was never meant to be API, it was (by mistake) missing the SPI annotation, so adding that here. While it's technically an API-break, nobody should have been using this method.

PR #70, opened 2023-11-08
Add README badges

czechboy0:
### Motivation

Surface the Swift version and platform support status from Swift Package Index.

### Modifications

Added badges, plus a quick link to the docc docs, to the top of the README.

### Result

Easier to quickly see our support matrix, plus the quick link to docs.

### Test Plan

Previewed locally.


README.md
```diff
@@ -1,5 +1,9 @@
 # Swift OpenAPI Generator Runtime
 
+[![](https://img.shields.io/badge/docc-read_documentation-blue)](https://swiftpackageindex.com/apple/swift-openapi-runtime/documentation)
```
MahdiBM, line 3: Would you want to add the SSWG badge too perhaps, or have you already thought about it?
With an `img` tag it would look like so:
```
<img src="https://user-images.githubusercontent.com/54685446/201329617-9fd91ab0-35c2-42c2-8963-47b68c6a490a.png" alt="SSWG sandbox">
```

MahdiBM: I see the generator repository does have it, which makes me think not including the badge here was intentional and was to just keep the badge to the main repo.

simonjbeaumont (approved):

PR #72, opened 2023-11-14
[Multipart] Introduce a bytes -> frames parser

czechboy0:
### Motivation

Start landing the parts of the multipart machinery that is unlikely to change as part of the multipart proposal that's finishing review tomorrow.

### Modifications

Introduce a bytes -> frames parser and an async sequence that wraps it.

A "frame" is either the full header fields section or a single chunk of a part body.

### Result

We can now frame bytes of a multipart body.

### Test Plan

Added unit tests for the state machine, the parser, and the async sequence.

### DIFF
--- a/Sources/OpenAPIRuntime/Multipart/ByteUtilities.swift
+++ b/Sources/OpenAPIRuntime/Multipart/ByteUtilities.swift
@@ -57,10 +57,10 @@ enum ASCII {
     }
 }
 
-/// A value returned by the `firstIndexAfterElements` method.
-enum FirstIndexAfterElementsResult<C: RandomAccessCollection> {
+/// A value returned by the `firstIndexAfterPrefix` method.
+enum FirstIndexAfterPrefixResult<C: RandomAccessCollection> {
 
-    /// The index after the end of the first match.
+    /// The index after the end of the prefix match.
     case index(C.Index)
 
     /// Matched all characters so far, but reached the end of self before matching all.
@@ -68,7 +68,7 @@ enum FirstIndexAfterElementsResult<C: RandomAccessCollection> {
     case reachedEndOfSelf
 
     /// The character at the provided index does not match the expected character.
-    case mismatchedCharacter(C.Index)
+    case unexpectedPrefix(C.Index)
 }
 
 extension RandomAccessCollection where Element: Equatable {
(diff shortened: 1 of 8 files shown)


Sources/OpenAPIRuntime/Multipart/ByteUtilities.swift
```diff
@@ -0,0 +1,121 @@
...
+    static let lf: UInt8 = 0x0a
+
+    /// The colon `:` character.
+    static let colon: UInt8 = 0x3a
+
+    /// The space ` ` character.
+    static let space: UInt8 = 0x20
+
+    /// The horizontal tab `<TAB>` character.
+    static let tab: UInt8 = 0x20
```
simonjbeaumont, line 34:
```suggestion
    static let tab: UInt8 = 0x09
```

Sources/OpenAPIRuntime/Multipart/MultipartParser.swift
```diff
@@ -0,0 +1,342 @@
...
+                            .firstIndex(where: { !ASCII.isValidHeaderFieldNameByte($0) })
+                    else {
+                        // No index matched yet, we need more data.
+                        state = .parsingPart(buffer, .parsingHeaderFields(headerFields))
+                        return .needsMore
+                    }
+                    let startHeaderValueWithWhitespaceIndex: [UInt8].Index
+                    // Check that what follows is a colon, otherwise this is a malformed header field line.
+                    // Source: RFC 7230, section 3.2.4.
+                    switch buffer[endHeaderNameIndex...].firstIndexAfterElements([ASCII.colon]) {
```
simonjbeaumont, line 234: ~I assume this is a check for `No whitespace is allowed between the header field-name and colon.`, however, this just finds the first index after the next colon and if it finds one, then it sets that as the start of the header value. FWICT it only returns `.emitError(.missingColonAfterHeaderName)` if there is no colon.~

I wrote all of that with a false assumption of what `firstIndexAfterElements` does, which, I think could do with renaming. From the name alone, I assumed it would return the first index after the elements if they exist anywhere in the remaining sequence, not just at the start.

simonjbeaumont (approved):
Looks really great to me. Only one blocking comment (a typo in the character codes).

Sources/OpenAPIRuntime/Multipart/ByteUtilities.swift
czechboy0: Copy pasta FTL.

Thanks for catching that.

Sources/OpenAPIRuntime/Multipart/MultipartParser.swift
czechboy0: Good point, ideas for a better name?

simonjbeaumont: It looks like this is usually followed by removing these elements from the front of the buffer. Do you think we could express this in some form of the term "pop", e.g. `popPrefix`. Maybe it would be mutating and return an enum value, simliar to how you do now. If the prefix _is_ there it is returned (and dropped from the collection), if it isn't you get either `.reachedEnd` or `.unexpectedPrefix`.

Then you can progress through the bytes that you have with e.g. `buffer.popPrefix([colon])`

czechboy0: There are cases where we keep reading more bytes before dropping the prefix, so I wouldn't turn it into a mutating function, but I like your suggestion around the "prefix" terminology - that's much better. Let me update it.

czechboy0: Updated: https://github.com/apple/swift-openapi-runtime/pull/72/commits/115734250be5359adc7e617e471e28f6a3cf1de8

PR #73, opened 2023-11-15
[Multipart] Add the frame -> bytes serializer

czechboy0:
### Motivation

Second state machine, doing the inverse of #72, serializes frames into bytes.

### Modifications

- A new state machine.
- A new serializer wrapping the state machine.
- An async sequence wrapping the serializer.

### Result

We can now serialize multipart frames into bytes.

### Test Plan

Unit tests for all 3 layers.


Sources/OpenAPIRuntime/Multipart/MultipartFramesToBytesSequence.swift
```diff
@@ -0,0 +1,71 @@
...
+    var upstream: Upstream
+
+    /// The boundary string used to separate multipart parts.
+    var boundary: String
+}
+
+extension MultipartFramesToBytesSequence: AsyncSequence {
+
+    /// The type of element produced by this asynchronous sequence.
+    typealias Element = ArraySlice<UInt8>
```
simonjbeaumont, line 31: How do you feel about making use of the byte chunk here.

```suggestion
    typealias Element = HTTPBody.ByteChunk
```

Sources/OpenAPIRuntime/Multipart/MultipartSerializer.swift
```diff
@@ -0,0 +1,260 @@
...
+    struct StateMachine {
+
+        /// The possible states of the state machine.
+        enum State: Hashable {
+
+            /// Has not yet written any bytes.
+            case initial
+
+            /// Emitted start, but no frames yet.
+            case startedNothingEmittedYet
```
simonjbeaumont, line 148: Needs a better name since the comment suggests emitted _something_ (cf. nothing); specifically the start.

For symmetry with the others, why not, simply `emittedStart`?

simonjbeaumont (approved):
Looks good. Minor comment.

Sources/OpenAPIRuntime/Multipart/MultipartFramesToBytesSequence.swift
czechboy0: Yeah I had that originally, but then I found it nicer in tests to not couple it to HTTPBody in any way.

Sources/OpenAPIRuntime/Multipart/MultipartSerializer.swift
czechboy0: Sure, good call.

czechboy0: https://github.com/apple/swift-openapi-runtime/pull/73/commits/786d00e0dbf1ab040fd59e7e9286bb85b5d193cc

PR #74, opened 2023-11-16
[Multipart] Add a frames -> raw parts parsing sequence

czechboy0:
### Motivation

Next sequence, parses `MultipartRawPart` values from a sequence of `MultipartFrame` values.

### Modifications

Moved the MultipartParser/Serializer into the files where their respective sequences live, it's better organized this way. But it's just a move so doesn't need to be re-reviewed.

This sequence is a bit unique, as this is where we need to handle the two sequences pulling from a common source of truth.

On the consumer side, this allows you to do:

```swift
for try await part in partsSequence {
    for try await chunk in part.body {
        // process chunk of this part
    }
}
```

The outer sequence has the element of `MultipartRawType`, which has header fields, and an `HTTPBody` (which itself is an async sequence, of byte chunks). This sequence vends `MultipartRawType` values, and needs to demultiplex the frames appropriately to the right async sequence (either the outer one vending parts, or the inner body one of the last-vended part).

It is a requirement that the part's body chunks (inner sequence) must be fully consumed before the outer sequence is asked for the next part. The state machine enforces this.

The solution is by using an actor that both iterators (outer and each of the inner ones) have a reference to and can ask for frames.

Also, had to disable strict concurrency checking on 5.10, because they added a warning when you iterate an async sequence from an actor; it's known, no workaround AFAIK. The alternative here would be to turn off warnings-as-errors for 5.10.

### Result

A parsing sequence that turns a sequence of frames into a sequence of parts, each of which has a nested sequence of body chunks.

### Test Plan

Unit tests added at all 3 layers:
- sequence
- shared iterator actor
- state machine of the shared iterator


Sources/OpenAPIRuntime/Multipart/MultipartBytesToFramesSequence.swift
```diff
@@ -65,3 +66,338 @@ extension MultipartBytesToFramesSequence: AsyncSequence {
         mutating func next() async throws -> MultipartFrame? { try await parser.next { try await upstream.next() } }
     }
 }
+
```
czechboy0, line 69: This was just moved here from `MultipartParser`, no need to re-review.

Sources/OpenAPIRuntime/Multipart/MultipartFramesToBytesSequence.swift
```diff
@@ -69,3 +70,247 @@ extension MultipartFramesToBytesSequence: AsyncSequence {
         }
     }
 }
+
```
czechboy0, line 73: This was just moved here from `MultipartSerializer`, no need to re-review.

Sources/OpenAPIRuntime/Multipart/MultipartFramesToRawPartsSequence.swift
czechboy0, line 1: This is actually new, please review.

Tests/OpenAPIRuntimeTests/Multipart/Test_MultipartBytesToFramesSequence.swift
```diff
@@ -36,3 +36,146 @@ final class Test_MultipartBytesToFramesSequence: Test_Runtime {
         )
     }
 }
+
```
czechboy0, line 39: This was just moved here from `Test_MultipartParser`, no need to re-review.

Tests/OpenAPIRuntimeTests/Multipart/Test_MultipartFramesToBytesSequence.swift
```diff
@@ -34,3 +34,66 @@ final class Test_MultipartFramesToBytesSequence: Test_Runtime {
         XCTAssertEqualData(bytes, expectedBytes)
     }
 }
+
+final class Test_MultipartSerializer: Test_Runtime {
```
czechboy0, line 38: This was just moved here from `Test_MultipartSerializer`, no need to re-review.

Sources/OpenAPIRuntime/Multipart/MultipartFramesToRawPartsSequence.swift
```diff
@@ -0,0 +1,380 @@
...
+        /// - Throws: When a parsing error is encountered.
+        func nextFromPartSequence(bodyClosure: @escaping @Sendable () async throws -> ArraySlice<UInt8>?) async throws
+            -> Element?
+        {
+            switch stateMachine.nextFromPartSequence() {
+            case .returnNil: return nil
+            case .fetchFrame:
+                var upstream = upstream
+                let frame = try await upstream.next()
+                self.upstream = upstream
```
simonjbeaumont, line 347: This dance seems odd. Because of it being an `actor`?

simonjbeaumont (approved):
This looks good to me. Did you want more eyes on the state machine logic?

Sources/OpenAPIRuntime/Multipart/MultipartFramesToRawPartsSequence.swift
czechboy0: Yes, this is one of the issues when using async sequences from actors.

Weird indeed.

PR #79, opened 2023-11-27
Change type of HTTPBody.Length.known from Int to Int64

simonjbeaumont:
### Motivation

The associated value of  enum case `HTTPBody.Length.known` was `Int`, and, on watchOS, `Int = Int32`, which is not ideal for a content length.

### Modifications

- (API breaking) Change type of HTTPBody.Length.known from Int to Int64.

### Result

Can now express larger values for content type on 32-bit platforms.

### Test Plan

Unit tests pass.

### Related Issues

- Fixes https://github.com/apple/swift-openapi-generator/issues/354.

### Note

I have marked this PR as `semver/major` as it constitutes the first of the PRs we make in the run up to 1.0.0-alpha.1 and the release notes generator should therefore catch us from inadvertently cutting another pre-1.0 patch release.

### DIFF
--- a/Sources/OpenAPIRuntime/Interface/HTTPBody.swift
+++ b/Sources/OpenAPIRuntime/Interface/HTTPBody.swift
@@ -125,17 +125,17 @@ public final class HTTPBody: @unchecked Sendable {
     /// the input sequence can be iterated.
     public let iterationBehavior: OpenAPIRuntime.IterationBehavior
 
-    /// Describes the total length of the body, if known.
+    /// Describes the total length of the body, in bytes, if known.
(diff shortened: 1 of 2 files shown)


simonjbeaumont:
@czechboy0 we discussed previously on the Github issue renaming this to `ContentLength` and `contentLength` to allow for a staged upgrade path, but given we are now about to break API for 1.0.0-alpha, I did not do this. Happy to reconsider. WDYT?

czechboy0:
No let's just change the type, no need to stage this in. We'll be making more breaking changes before 1.0.0-alpha.1 anyway.

czechboy0 (approved):

PR #88, opened 2023-12-04
[Docs] Make sure all symbols are curated

czechboy0:
### Motivation

Ensure all symbols are curated.

### Modifications

Added symbols that were showing up in automatic sections to existing curated sections.

### Result

All symbols are curated.

### Test Plan

Built docs locally.

### DIFF
--- a/Sources/OpenAPIRuntime/Documentation.docc/Documentation.md
+++ b/Sources/OpenAPIRuntime/Documentation.docc/Documentation.md
@@ -68,6 +68,18 @@ Please report any issues related to this library in the [swift-openapi-generator
 - ``Configuration``
 - ``DateTranscoder``
 - ``ISO8601DateTranscoder``
+- ``MultipartBoundaryGenerator``
+- ``RandomMultipartBoundaryGenerator``
+- ``ConstantMultipartBoundaryGenerator``
+- ``IterationBehavior``
+
+### Content types
+- ``HTTPBody``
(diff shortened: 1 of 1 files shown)


Sources/OpenAPIRuntime/Documentation.docc/Documentation.md
```diff
@@ -68,15 +68,29 @@ Please report any issues related to this library in the [swift-openapi-generator
...
+- ``MultipartPart``
+- ``MultipartDynamicallyNamedPart``
 
 ### Errors
 - ``ClientError``
 - ``ServerError``
 - ``UndocumentedPayload``
 
 ### HTTP Currency Types
-- ``HTTPBody``
```
simonjbeaumont, line 78: IMO I'd keep HTTPBody here.

simonjbeaumont (approved):
LGTM, but added a comment for you to consider about the categorisation of HTTPBody. IMO, this, by definition, is a HTTP currency type

Sources/OpenAPIRuntime/Documentation.docc/Documentation.md
czechboy0: It serves both as the HTTP currency type and also as the concrete type for raw bytes and plain text bodies, so I think it also makes sense under content types.

Added to both, here it makes sense for the type to be curated under two sections, to reflect the two places people might encounter it.

PR #90, opened 2023-12-14
[Runtime] Add headerFields and body to UndocumentedPayload

czechboy0:
### Motivation

The runtime changes for https://github.com/apple/swift-openapi-generator/issues/299.

### Modifications

Added `headerFields` and `body` properties to `UndocumentedPayload`, allowing adopters to access the full request information when an undocumented response is returned.

### Result

Easier access to the raw payload when an undocumented response is returned.

### Test Plan

These are just the runtime changes, tested together with generated changes.

### DIFF
--- a/Sources/OpenAPIRuntime/Deprecated/Deprecated.swift
+++ b/Sources/OpenAPIRuntime/Deprecated/Deprecated.swift
@@ -15,3 +15,10 @@ import Foundation
 import HTTPTypes
 
 // MARK: - Functionality to be removed in the future
+
+extension UndocumentedPayload {
+    /// Creates a new payload.
+    @available(*, deprecated, renamed: "init(headerFields:body:)") @_disfavoredOverload public init() {
+        self.init(headerFields: [:], body: nil)
+    }
+}
--- a/Sources/OpenAPIRuntime/Base/UndocumentedPayload.swift
+++ b/Sources/OpenAPIRuntime/Base/UndocumentedPayload.swift
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import HTTPTypes
+
 /// A payload value used by undocumented operation responses.
 ///
 /// Each operation's `Output` enum type needs to exhaustively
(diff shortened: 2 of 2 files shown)


Sources/OpenAPIRuntime/Deprecated/Deprecated.swift
```diff
@@ -15,3 +15,10 @@ import Foundation
 import HTTPTypes
 
 // MARK: - Functionality to be removed in the future
+
+extension UndocumentedPayload {
+    /// Creates a new payload.
+    @available(*, deprecated, renamed: "init(headerFields:body:)") @_disfavoredOverload public init() {
+        self.init(headerFields: [:], body: nil)
+    }
+}
```
simonjbeaumont, line 24: Do we need to carry this forever now?

Why do we need it, given the new `init(headerFields:body:)` has default parameters? 

Can't we do this differently by having the new `init(headerFields:body:)` _not_ have default parameters and just change the implementation of `init()` to call it with the defaults, given that we already have `init()` in the API?

Or, isn't it just enough to have `_disfavoredOverload` on the old `init()`?

All these questions are really just: can we avoid having a deprecated symbol here?

czechboy0: Unfortunately, we can't, because people could be capturing the initializer as a function.

Having the defaulted parameter means that people (and we, in tests) can do `UndocumentedPayload(headerFields: ...)` (no body), which is useful.

Basically, any change to any function arguments are technically source-breaking, so we always have to retain the previous one as well for the rest of 1.x. That said, I don't think it's a big deal.

simonjbeaumont: > Basically, any change to any function arguments are technically source-breaking, so we always have to retain the previous one as well for the rest of 1.x.

Understood. But I don't think either of my suggestions above were suggesting we remove it? Just that we keep it and forward calls and/or add `_disfavoredOverload`.

czechboy0: The above one has `_disfavoredOverload`, which ensures that we don't get warnings when you still use `UndocumentedPayload()`.

With this in mind, do you still believe it should change, and if so, how?

simonjbeaumont: > With this in mind, do you still believe it should change, and if so, how?

Forgive me if this is misguided, but I'm asking why we need to have the deprecated annotation?

Why not just keep them both in the same file. Concretely, keep both of these functions as they are now, i.e. with the `@_disfavoredOverload ` but _without_ the deprecation.

I don't see what the deprecation buys us at all. It's not like we don't want people to be able to call the function with no arguments. If we _did_ want to discourage calling it with no/default arguments then the new one shouldn't have default arguments. And, if the new one didn't have default arguments then we could keep the old one around anyway because it can forward with no ambiguity without even the use of `@_disfavoredOverload`.

czechboy0: Ah gotcha, I see what you mean now. Err, I marked it deprecated because if we were starting from scratch, this `init()` function would never exist, so for 2.0, I want to make sure we delete it, as it's unnecessary. The `init(headerFields:body:)` is all we need.

If we kept both, I suspect others (or us, in the future) would come back and not be sure why we have two overlapping initializer. The only reason is historical, because we shipped 1.0 with the `init()` one.

simonjbeaumont (approved):

PR #91, opened 2023-12-14
[Runtime] SOAR-0010: Event streams sequences

czechboy0:
### Motivation

Land changes approved in https://github.com/apple/swift-openapi-generator/pull/495.

### Modifications

Introduced the new APIs.

### Result

Easy use of event streams.

### Test Plan

Added unit tests for all.


Sources/OpenAPIRuntime/EventStreams/JSONSequenceDecoding.swift
```diff
@@ -0,0 +1,236 @@
...
+                state = .mutating
+                buffer.removeFirst()
+                state = .parsingLine(buffer: buffer)
+                return .noop
+            case .parsingLine(var buffer):
+                state = .mutating
+                guard let indexOfRecordSeparator = buffer.firstIndex(of: ASCII.rs) else {
+                    state = .parsingLine(buffer: buffer)
+                    return .needsMore
+                }
```
simonjbeaumont, line 180: Not a blocker, but I'm wondering if we can do better here with respect to performance. Specifically, this will call `firstIndex(of:)` every time we receive new bytes to find the record separator, which will be O(n) in the current size of the buffer and will search for the record separator through portions of the buffer that may have been searched last time.

A pathological example would be a sequence of one-byte chunks, in which case it will be quadratic in the size of the line, but it needn't be.

It can be linear if we:

1. Augment the `.parsingLine(buffer:)` state to store the current `endIndex` of the buffer before returning `.needsMore` or something similar.
2. Use this with a new utility function `firstIndex(of:from:)`, which would look for the first index from a starting index.

This comment might apply to the other sequences in this PR too.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
```diff
@@ -0,0 +1,451 @@
...
+                }
+                // Process the field.
+                switch field {
+                case "event": event.event = value
+                case "data":
+                    var data = event.data ?? ""
+                    data.append(value)
+                    data.append("\n")
+                    event.data = data
+                case "id": event.id = value
```
simonjbeaumont, line 207: Spec says:
> If the field value does not contain U+0000 NULL, then set the last event ID buffer to the field value. Otherwise, ignore the field.

Are we relying on earlier logic / string parsing for this?

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
```diff
@@ -0,0 +1,451 @@
...
+    ///
+    /// https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream
+    ///
+    /// This is not trivial to do with a streaming parser, as the end of line can be:
+    /// - LF
+    /// - CR
+    /// - CRLF
+    ///
+    /// So when we get CR, but have no more data, we want to be able to emit the previous line,
+    /// however we need to discard a LF if one comes.
```
simonjbeaumont, line 331: I think that we might be over reaching here and we can simplify even if it means delays. I.e. we can wait for the LF before emitting. It's covered explicitly by the spec:

> Since connections established to remote servers for such resources are expected to be long-lived, UAs should ensure that appropriate buffering is used. In particular, while line buffering with lines are defined to end with a single U+000A LINE FEED (LF) character is safe, block buffering or line buffering with different expected line endings can cause delays in event dispatch.

I read this as: if we can simplify here, we should. It's on the user to use good line endings.

Sources/OpenAPIRuntime/EventStreams/JSONSequenceDecoding.swift
czechboy0: It'd apply to other sequences _outside_ this PR as well, the multipart ones. Could you file that as an issue? I'm not closed to it and I even tried it briefly, but it exploded the complexity a bit too much for my liking, so I backed it out. But it was also during the prototyping phase, where everything was in flux, so I'd be happy to investigate doing this in all the parsers in the runtime library as a standalone piece of work.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
czechboy0: The specification describes the EventSource abstraction, a stateful representation of a stream, which helps you work with Server-sent Events.

My implementation only provides the Server-sent Events portion, but not the EventSource part. 

EventSource is currently up to adopters to implement to their liking, because it also contains actually making requests that establish the stream, which is out of scope here. In the future we could also implement EventSource it in the runtime library, but it wasn't a blocker here, which is why it was kept out.

The EventSource value has a "last event ID" field, so it's not relevant to the underlying SSE encoder/decoder.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
czechboy0: Are you proposing we don't support `<CR>`-delimited SSE at all? If so, that'd be the result - if we wait for a `<LF>` or end of stream before dispatching an event. You'd get all the events only after the server closed the connection at once, which basically means an unusable SSE stream.

Sources/OpenAPIRuntime/EventStreams/JSONSequenceDecoding.swift
simonjbeaumont: Filed an issue to track this separately: https://github.com/apple/swift-openapi-runtime/issues/93#issue-2070219477

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
simonjbeaumont: No, I was proposing that it's justified for us to simplify our logic such that the events are only eagerly emitted for LF-delimited streams as that was how I read this warning in the spec. Specifically, I think our logic is _correct_ but doing more than strictly necessary. If, given this understanding, we don't think this additional complexity is something we're desperate to shed then let's just keep it the way it is in this PR.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
simonjbeaumont: OK, thanks for confirming!

simonjbeaumont (approved):

AFutureD:
Hi @czechboy0 , I wonder why `ServerSentEvent` need to conform to `Hashable`?

czechboy0:
Hi @AFutureD - can you elaborate on what's behind that question? Is the Hashable conformance limiting your use of it?

AFutureD:
Hi @czechboy0 , Thanks for quick resposne.

In `ServerSentEventWithJSONData`, the generic JSONDataType may not guaranteed to conform to Hashable. 

For example, there are a lot of packages can be found on GitHub for `JsonSchema`, and they all do not conform to Hashable.

And, the decoder or deserializer may not need Hashable constraint.

PR #92, opened 2023-12-20
Mark RuntimeError as @_spi(Generated) too

MrSkwiggs:
### Motivation

I'm writing a networking library that makes use of the heavy lifting provided by `UniversalClient`, and figured I'd use `RuntimeError` as well since they cover pretty much all of my use-cases. 

### Modifications

Marked `RuntimeError` as `@_spi(Generated) public` instead of `internal`

### Result

`RuntimeError` can be imported using `@_spi(Generated) import enum OpenAPIRuntime.RuntimeError`

### Test Plan

None

### DIFF
--- a/Sources/OpenAPIRuntime/Errors/RuntimeError.swift
+++ b/Sources/OpenAPIRuntime/Errors/RuntimeError.swift
@@ -15,7 +15,7 @@ import protocol Foundation.LocalizedError
 import struct Foundation.Data
 
 /// Error thrown by generated code.
-internal enum RuntimeError: Error, CustomStringConvertible, LocalizedError, PrettyStringConvertible {
+@_spi(Generated) public enum RuntimeError: Error, CustomStringConvertible, LocalizedError, PrettyStringConvertible {
 
     // Miscs
     case invalidServerURL(String)


czechboy0:
Hi @MrSkwiggs, `RuntimeError` is deliberately not API (not even SPI), because it's an enum and for those, any additional case counts as an API-breaking change.

Can you share more about your use case? Maybe we can find another way to help you get what you need.

MrSkwiggs:
Heya @czechboy0 👋 

Well, I am really just writing a networking library by hand, not using OpenAPI-generated code (the endpoint I'm consuming doesn't provide any OpenAPI scheme...). 

Not wanting to reinvent the wheel, I did see a lot of the work provided by the Universal client was did what I needed, as well as the `RuntimeError` part (especially when writing default conformances, i.e I have a `Request` protocol that I use in a bridging function to `UniversalClient`'s `send(_:)` function).

Things such as:

```swift
protocol Request {
    associatedtype Response: Sendable, Decodable

    var route: Route { get }

    var method: HTTPRequest.Method { get }

    func serialize() throws -> (HTTPRequest, HTTPBody?)
    func deserialize(response: HTTPResponse, body: HTTPBody?) async throws -> Response
}

// MARK: GET
protocol GetRequest: Request {}
extension GetRequest {
    var method: HTTPRequest.Method { .get }
    
    func serialize() throws -> (HTTPRequest, HTTPBody?) {
        return (
            HTTPRequest(soar_path: route.path, method: method),
            nil
        )
    }

    func deserialize(response: HTTPResponse, body: HTTPBody?, using decoder: Decoder) async throws -> Response {
        switch response.status.code {
        case 200:
            guard let body else { throw RuntimeError.missingRequiredResponseBody }
            let dataBuffer = try await ArraySlice(collecting: body, upTo: .max)
            return try JSONDecoder().decode(Response.self, from: Data(dataBuffer))
        default:
            throw RuntimeError.unexpectedResponseStatus(expectedStatus: "200", response: response)
        }
    }
}
```

and

```swift
extension UniversalClient {
    func send<T: GetRequest>(request: T) async throws -> T.Response {
        try await send(
            input: (),
            forOperation: request.identifier) { _ in
                try request.serialize()
            } deserializer: { response, body in
                try await request.deserialize(response: response, body: body)
            }
    }
}
```

It's a small use-case, but essentially I find it very convenient to throw `RuntimeError`s in my default conformances

czechboy0:
I'd strongly recommend you do _not_ use any of our `@_spi(Generated)` APIs directly, we reserve the right to break them without warning as long as we make sure the generator itself (the only consumer of these APIs) can handle it. You will not have a good time.

If you're interested in the HTTP types, those aren't coming from this library, but from https://github.com/apple/swift-http-types. If you're looking for an HTTP client, and URLSession doesn't fit your needs, you could consider https://github.com/swift-server/async-http-client.

But if you're not using OpenAPI, I don't think you'll get much value using this library.

MahdiBM:
@MrSkwiggs 

> the endpoint I'm consuming doesn't provide any OpenAPI scheme...)

It could be worth it to create an openapi document for that API yourself.
Then you can use this library to generate code for the endpoints like normal.

PR #94, opened 2024-01-14
Make the ISO8601 date transcoder configurable

czechboy0:
### Motivation

Fixes #389.

### Modifications

Add a way to configure the existing `ISO8601DateTranscoder`.

### Result

Adopters can more easily use e.g. fractional seconds-based transcoder.

### Test Plan

Added unit tests.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/Configuration.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Configuration.swift
@@ -25,14 +25,45 @@ public protocol DateTranscoder: Sendable {
 }
 
 /// A transcoder for dates encoded as an ISO-8601 string (in RFC 3339 format).
-public struct ISO8601DateTranscoder: DateTranscoder {
+public struct ISO8601DateTranscoder: DateTranscoder, @unchecked Sendable {
+
+    /// The lock protecting the formatter.
+    private let lock: NSLock
+
+    /// The underlying date formatter.
+    private let locked_formatter: ISO8601DateFormatter
+
+    /// Creates a new default transcoder.
+    public init() { self.init(optionsOverride: nil) }
+
+    /// Creates a new transcoder with the provided options.
+    /// - Parameter optionsOverride: Options to override the default ones. If you provide
+    ///   nil here, the default options are used.
+    public init(optionsOverride: ISO8601DateFormatter.Options?) {
+        let formatter = ISO8601DateFormatter()
+        if let optionsOverride { formatter.formatOptions = optionsOverride }
+        self.init(formatter: formatter)
+    }
+
+    /// Creates a new transcoder with the provided formatter.
+    /// - Parameter formatter: The formatter to use.
+    public init(formatter: ISO8601DateFormatter) {
+        lock = NSLock()
+        locked_formatter = formatter
+    }
 
     /// Creates and returns an ISO 8601 formatted string representation of the specified date.
-    public func encode(_ date: Date) throws -> String { ISO8601DateFormatter().string(from: date) }
+    public func encode(_ date: Date) throws -> String {
+        lock.lock()
+        defer { lock.unlock() }
+        return locked_formatter.string(from: date)
+    }
 
     /// Creates and returns a date object from the specified ISO 8601 formatted string representation.
     public func decode(_ dateString: String) throws -> Date {
-        guard let date = ISO8601DateFormatter().date(from: dateString) else {
+        lock.lock()
+        defer { lock.unlock() }
+        guard let date = locked_formatter.date(from: dateString) else {
             throw DecodingError.dataCorrupted(
                 .init(codingPath: [], debugDescription: "Expected date string to be ISO8601-formatted.")
             )
@@ -44,6 +75,11 @@ public struct ISO8601DateTranscoder: DateTranscoder {
 extension DateTranscoder where Self == ISO8601DateTranscoder {
     /// A transcoder that transcodes dates as ISO-8601–formatted string (in RFC 3339 format).
     public static var iso8601: Self { ISO8601DateTranscoder() }
+
+    /// A transcoder that transcodes dates as ISO-8601–formatted string (in RFC 3339 format) with fractional seconds.
+    public static var iso8601WithFractionalSeconds: Self {
+        ISO8601DateTranscoder(optionsOverride: [.withInternetDateTime, .withFractionalSeconds])
+    }
 }
 
 extension JSONEncoder.DateEncodingStrategy {
--- a/Tests/OpenAPIRuntimeTests/Conversion/Test_Configuration.swift
+++ b/Tests/OpenAPIRuntimeTests/Conversion/Test_Configuration.swift
@@ -1,0 +1,30 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftOpenAPIGenerator project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import XCTest
+@_spi(Generated) import OpenAPIRuntime
+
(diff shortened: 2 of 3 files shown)


Sources/OpenAPIRuntime/Conversion/Configuration.swift
```diff
@@ -25,14 +25,45 @@ public protocol DateTranscoder: Sendable {
...
+    /// The underlying date formatter.
+    private let locked_formatter: ISO8601DateFormatter
+
+    /// Creates a new default transcoder.
+    public init() { self.init(optionsOverride: nil) }
+
+    /// Creates a new transcoder with the provided options.
+    /// - Parameter optionsOverride: Options to override the default ones. If you provide
+    ///   nil here, the default options are used.
+    public init(optionsOverride: ISO8601DateFormatter.Options?) {
```
czechboy0, line 42: Looking for feedback on this parameter here. Not happy with it yet.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
```diff
@@ -25,14 +25,45 @@ public protocol DateTranscoder: Sendable {
...
+        let formatter = ISO8601DateFormatter()
+        if let optionsOverride { formatter.formatOptions = optionsOverride }
+        self.init(formatter: formatter)
+    }
+
+    /// Creates a new transcoder with the provided formatter.
+    /// - Parameter formatter: The formatter to use.
+    public init(formatter: ISO8601DateFormatter) {
```
czechboy0, line 50: Do we need this one? Technically, the formatter is an open class, so someone could have a subclass they'd want to provide.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
```diff
@@ -44,6 +75,11 @@ public struct ISO8601DateTranscoder: DateTranscoder {
 extension DateTranscoder where Self == ISO8601DateTranscoder {
     /// A transcoder that transcodes dates as ISO-8601–formatted string (in RFC 3339 format).
     public static var iso8601: Self { ISO8601DateTranscoder() }
+
+    /// A transcoder that transcodes dates as ISO-8601–formatted string (in RFC 3339 format) with fractional seconds.
+    public static var iso8601WithFractionalSeconds: Self {
```
czechboy0, line 80: Looking for naming feedback here.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
simonjbeaumont: Maybe if we spell this as `init(options: ISO8601DateFormatter.Options? = nil)`, the optionality and nil-default will imply a default-when-nil, which can also be documented for additional clarity.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
```diff
@@ -25,14 +25,45 @@ public protocol DateTranscoder: Sendable {
...
 
 /// A transcoder for dates encoded as an ISO-8601 string (in RFC 3339 format).
-public struct ISO8601DateTranscoder: DateTranscoder {
+public struct ISO8601DateTranscoder: DateTranscoder, @unchecked Sendable {
+
+    /// The lock protecting the formatter.
+    private let lock: NSLock
+
+    /// The underlying date formatter.
+    private let locked_formatter: ISO8601DateFormatter
```
simonjbeaumont, line 34: Do we need to have a stored property for the formatter and consequently have to deal with the locking? Could we instead hold an `IOS8601DateFormatter.Options` and use this when we create the formatter? 

Alternatively, I notice that `ISO8601DateFormatter` has [`class func string(from:timeZone:formatOptions:)`](https://developer.apple.com/documentation/foundation/iso8601dateformatter/1642834-string) so maybe we just use that and pass the options that the user provides?

Sources/OpenAPIRuntime/Conversion/Configuration.swift
simonjbeaumont: If we don't need it now, let's drop it until we have a need.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
```diff
@@ -25,14 +25,45 @@ public protocol DateTranscoder: Sendable {
...
+    public init(formatter: ISO8601DateFormatter) {
+        lock = NSLock()
+        locked_formatter = formatter
+    }
 
     /// Creates and returns an ISO 8601 formatted string representation of the specified date.
-    public func encode(_ date: Date) throws -> String { ISO8601DateFormatter().string(from: date) }
+    public func encode(_ date: Date) throws -> String {
+        lock.lock()
+        defer { lock.unlock() }
```
simonjbeaumont, line 58: If we do decide to keep the formatter and still need locking can we pull in the `LockedValueBox` so we don't need to use unchecked sendable?

Sources/OpenAPIRuntime/Conversion/Configuration.swift
simonjbeaumont: Maybe we don't have it, but if we do I think the name is clear enough. OOI, is it a problem to include fractional seconds in `.iso8601`?

Sources/OpenAPIRuntime/Conversion/Configuration.swift
czechboy0: Performance is a concern here, but we mostly lived with it in the past. Now that we want to allow the adopter to provide a custom `ISO8601DateFormatter` (possibly even a subclass), which isn't `Sendable`, we need to lock around it somehow.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
czechboy0: Yeah I like that better, thanks.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
czechboy0: It's a spectrum, technically even customizing the options isn't needed, as adopters can implement a whole new date transcoder and have full flexibility. It's just been a common ask.

I think the question of this initializer is kind of related to the locking around the value. Do we create a new formatter under the hood for every call, or do we cache one and protect it with a lock? I haven't measured the performance difference here, maybe I should to inform this discussion more.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
czechboy0: Sure, we use NSLock in other places as well so we should probably migrate them all to LockedValueBox if we introduce it to the runtime library.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
simonjbeaumont: > Now that we want to allow the adopter to provide a custom ISO8601DateFormatter ...

I'm gently pushing back on whether that's a requirement. And whether it would be enough for them to provide just the formatter options.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
simonjbeaumont: Ah, that's sad. I'd have expected that the with fractional seconds would ensure they are present when printing, but be permissive to them not being present when parsing.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
czechboy0: Yeah, probably. Let's see what the perf test says so that we can balance these.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
czechboy0: That would be ideal, agreed.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
simonjbeaumont: Tracked by https://github.com/apple/swift-openapi-generator/issues/505.

simonjbeaumont (approved):

nmtitov:
@czechboy0 how to actually use new transcoder in projects?

nmtitov:
@czechboy0 thank you!

PR #95, opened 2024-01-18
Bug/502 crash thread safety fix

LarsPetersHH:
### Motivation

Fixes https://github.com/apple/swift-openapi-generator/issues/502
- Ensure thread safety of `HTTPBody.collect(upTo)`.
- `makeAsyncIterator()`: Instead of crashing, return AsyncSequence which throws `TooManyIterationsError` thereby honoring the contract for `IterationBehavior.single` (HTTPBody, MultipartBody)

### Modifications

- HTTPBody, MultipartBody: `makeAsyncIterator()`: removed `try!`, catch error and create a sequence which throws the error on iteration.
- This removed the need for `try checkIfCanCreateIterator()` in `HTTPBody.collect(upTo)`. 
**Note**: This creates a small change in behavior: There may be a `TooManyBytesError` thrown before the check for `iterationBehavior`. This approach uses the simplest code, IMO. If we want to keep that `iterationBehavior` is checked first and only after that for the length, then the code needs to be more complex.
- Removed `try checkIfCanCreateIterator()` in both classes (only used in `HTTPBody`).

### Result

- No intentional crash in `makeAsyncIterator()` anymore.
- Tests supplied as example in https://github.com/apple/swift-openapi-generator/issues/502 succeed.

### Test Plan

- Added check in `Test_Body.testIterationBehavior_single()` to ensure that using `makeAsyncIterator()` directly yields the expected error.
- Added tests to check iteration behavior of `MultipartBody`.

### DIFF
--- a/Sources/OpenAPIRuntime/Interface/HTTPBody.swift
+++ b/Sources/OpenAPIRuntime/Interface/HTTPBody.swift
@@ -336,9 +336,7 @@ extension HTTPBody: AsyncSequence {
         do {
             try tryToMarkIteratorCreated()
             return .init(sequence.makeAsyncIterator())
-        } catch {
-            return .init(throwing: error)
-        }
+        } catch { return .init(throwing: error) }
     }
 }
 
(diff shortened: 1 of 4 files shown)


czechboy0 (approved):
This is great, thank you!

PR #99, opened 2024-02-04
Add a hook for intercepting decoding errors

jpsim:
### Motivation

It's useful for clients to be able to have a centralized way to handle decoding errors, which can be useful for logging or telemetry.

See https://github.com/apple/swift-openapi-generator/issues/522 for more discussion on this.

### Modifications

Add a new `DecodingErrorHandler` protocol which has a single `willThrow(_ error: any Error)` requirement.

A decoding handler can be passed to the client configuration on initialization.

This is just a proof-of-concept, if this direction is reasonable, it should be fleshed out to handle all decoding paths.

### Result

With this change, errors produced by body decoders can be intercepted in a centralized way for all methods going through the `Client`.

### Test Plan

I added a simple unit test. More tests would need to be added to sufficiently cover the functionality being introduced here.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/Configuration.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Configuration.swift
@@ -96,6 +96,10 @@ extension JSONDecoder.DateDecodingStrategy {
     }
 }
 
+public protocol DecodingErrorHandler: Sendable {
+    func willThrow(_ error: any Error)
+}
+
 /// A set of configuration values used by the generated client and server types.
 public struct Configuration: Sendable {
 
@@ -105,6 +109,8 @@ public struct Configuration: Sendable {
     /// The generator to use when creating mutlipart bodies.
     public var multipartBoundaryGenerator: any MultipartBoundaryGenerator
 
+    public var decodingErrorHandler: (any DecodingErrorHandler)?
+
     /// Creates a new configuration with the specified values.
     ///
     /// - Parameters:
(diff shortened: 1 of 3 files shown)


Sources/OpenAPIRuntime/Conversion/Configuration.swift
```diff
@@ -96,6 +96,10 @@ extension JSONDecoder.DateDecodingStrategy {
     }
 }
 
+public protocol DecodingErrorHandler: Sendable {
+    func willThrow(_ error: any Error)
```
czechboy0, line 100: Would you be interested in the context this was thrown in? See `ClientError` and `ServerError`, where we add more context.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
```diff
@@ -113,9 +119,11 @@ public struct Configuration: Sendable {
     ///   - multipartBoundaryGenerator: The generator to use when creating mutlipart bodies.
     public init(
         dateTranscoder: any DateTranscoder = .iso8601,
-        multipartBoundaryGenerator: any MultipartBoundaryGenerator = .random
+        multipartBoundaryGenerator: any MultipartBoundaryGenerator = .random,
+        decodingErrorHandler: (any DecodingErrorHandler)? = nil
```
czechboy0, line 123: Looks good, please just add a deprecated 2-parameter version of this initializer in Deprecated.swift, as adding a param is technically a source break. Similar to https://github.com/apple/swift-openapi-runtime/pull/102#discussion_r1547909598

czechboy0 (commented):
Looks great overall, apologies for the delay on this PR, it somehow slipped through unnoticed.

czechboy0:
Hi @jpsim - are you still interested in a solution here? I'm happy with either you picking this up again, or letting us take over. It'd just be good to finalize the details with a concrete use case in mind.

czechboy0 (changes requested):
Marking as waiting on @jpsim for now

jpsim:
I won't be pursuing this further.

czechboy0:
> I won't be pursuing this further. 

May I ask if it's because you found a workflow where you don't need it, or if you don't have the bandwidth to drive the PR?

daviduzumeri:
We're using this library pretty heavily and this would be a fantastic addition that would help us catch a lot of stuff in Datadog instead of finding out API responses aren't decoding properly from customer complaints.

czechboy0:
We're open to this enhancement, and with JP closing this PR, I think it'd be appropriate for someone else to open a new PR and we can discuss next steps there.

PR #100, opened 2024-03-07
[Runtime] Add support of deepObject style in query params

kstefanou52:
### Motivation

The runtime changes for:
https://github.com/apple/swift-openapi-generator/issues/259

### Modifications

Added `deepObject` style to serializer & parser in order to support nested keys on query parameters.

### Result

Support nested keys on query parameters.

### Test Plan

These are just the runtime changes, tested together with generated changes.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/ParameterStyles.swift
+++ b/Sources/OpenAPIRuntime/Conversion/ParameterStyles.swift
@@ -26,6 +26,11 @@
     ///
     /// Details: https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.2
     case simple
+    
+    /// The deepObject style.
+    ///
+    /// Details: https://spec.openapis.org/oas/v3.1.0.html#style-values
+    case deepObject
 }
 
 extension ParameterStyle {
@@ -53,6 +58,7 @@ extension URICoderConfiguration.Style {
         switch style {
         case .form: self = .form
         case .simple: self = .simple
+        case .deepObject: self = .deepObject
         }
     }
 }
--- a/Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
+++ b/Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
@@ -43,6 +43,9 @@ private enum ParsingError: Swift.Error {
 
     /// A malformed key-value pair was detected.
     case malformedKeyValuePair(Raw)
+    
+    /// An invalid configuration was detected.
+    case invalidConfiguration(String)
 }
 
 // MARK: - Parser implementations
@@ -61,6 +64,7 @@ extension URIParser {
             switch configuration.style {
             case .form: return [:]
             case .simple: return ["": [""]]
+            case .deepObject: return [:]
             }
         }
         switch (configuration.style, configuration.explode) {
@@ -68,6 +72,10 @@ extension URIParser {
         case (.form, false): return try parseUnexplodedFormRoot()
         case (.simple, true): return try parseExplodedSimpleRoot()
         case (.simple, false): return try parseUnexplodedSimpleRoot()
+        case (.deepObject, true): return try parseExplodedDeepObjectRoot()
+        case (.deepObject, false):
+            let reason = "Deep object style is only valid with explode set to true"
+            throw ParsingError.invalidConfiguration(reason)
         }
     }
 
@@ -202,6 +210,43 @@ extension URIParser {
             while !data.isEmpty {
                 let value = data.parseUpToCharacterOrEnd(pairSeparator)
                 appendPair(.init(), [value])
+            }
+        }
+    }
+    
+    /// Parses the root node assuming the raw string uses the deepObject style
+    /// and the explode parameter is enabled.
+    /// - Returns: The parsed root node.
+    /// - Throws: An error if parsing fails.
+    private mutating func parseExplodedDeepObjectRoot() throws -> URIParsedNode {
+        try parseGenericRoot { data, appendPair in
+            let keyValueSeparator: Character = "="
+            let pairSeparator: Character = "&"
+            let nestedKeyStartingCharacter: Character = "["
+            let nestedKeyEndingCharacter: Character = "]"
+            
+            func nestedKey(from deepObjectKey: String.SubSequence) -> Raw {
+                var unescapedDeepObjectKey = Substring(deepObjectKey.removingPercentEncoding ?? "")
+                let topLevelKey = unescapedDeepObjectKey.parseUpToCharacterOrEnd(nestedKeyStartingCharacter)
+                let nestedKey = unescapedDeepObjectKey.parseUpToCharacterOrEnd(nestedKeyEndingCharacter)
+                return nestedKey.isEmpty ? topLevelKey : nestedKey
+            }
+            
+            while !data.isEmpty {
+                let (firstResult, firstValue) = data.parseUpToEitherCharacterOrEnd(
+                    first: keyValueSeparator,
+                    second: pairSeparator
+                )
+                
+                guard case .foundFirst = firstResult else {
+                    throw ParsingError.malformedKeyValuePair(firstValue)
+                }
+                // Hit the key/value separator, so a value will follow.
+                let secondValue = data.parseUpToCharacterOrEnd(pairSeparator)
+                let key = nestedKey(from: firstValue)
+                let value = secondValue
+                
+                appendPair(key, [value])
             }
         }
     }
--- a/Sources/OpenAPIRuntime/URICoder/Serialization/URISerializer.swift
+++ b/Sources/OpenAPIRuntime/URICoder/Serialization/URISerializer.swift
@@ -69,6 +69,12 @@ extension URISerializer {
 
         /// Nested containers are not supported.
         case nestedContainersNotSupported
+        
+        /// Deep object arrays are not supported.
+        case deepObjectsArrayNotSupported
+        
+        /// An invalid configuration was detected.
+        case invalidConfiguration(String)
     }
 
     /// Computes an escaped version of the provided string.
@@ -117,6 +123,7 @@ extension URISerializer {
             switch configuration.style {
             case .form: keyAndValueSeparator = "="
             case .simple: keyAndValueSeparator = nil
+            case .deepObject: keyAndValueSeparator = "="
             }
             try serializePrimitiveKeyValuePair(primitive, forKey: key, separator: keyAndValueSeparator)
         case .array(let array): try serializeArray(array.map(unwrapPrimitiveValue), forKey: key)
@@ -180,6 +187,8 @@ extension URISerializer {
         case (.simple, _):
             keyAndValueSeparator = nil
             pairSeparator = ","
+        case (.deepObject, _):
+            throw SerializationError.deepObjectsArrayNotSupported
         }
         func serializeNext(_ element: URIEncodedNode.Primitive) throws {
             if let keyAndValueSeparator {
@@ -228,8 +237,18 @@ extension URISerializer {
         case (.simple, false):
             keyAndValueSeparator = ","
             pairSeparator = ","
+        case (.deepObject, true):
+            keyAndValueSeparator = "="
+            pairSeparator = "&"
+        case (.deepObject, false):
+            let reason = "Deep object style is only valid with explode set to true"
+            throw SerializationError.invalidConfiguration(reason)
         }
 
+        func serializeNestedKey(_ elementKey: String, forKey rootKey: String) -> String {
+            guard case .deepObject = configuration.style else { return elementKey }
+            return rootKey + "[" + elementKey + "]"
+        }
         func serializeNext(_ element: URIEncodedNode.Primitive, forKey elementKey: String) throws {
             try serializePrimitiveKeyValuePair(element, forKey: elementKey, separator: keyAndValueSeparator)
         }
@@ -238,10 +257,13 @@ extension URISerializer {
             data.append(containerKeyAndValue)
         }
         for (elementKey, element) in sortedDictionary.dropLast() {
-            try serializeNext(element, forKey: elementKey)
+            try serializeNext(element, forKey: serializeNestedKey(elementKey, forKey: key))
             data.append(pairSeparator)
         }
-        if let (elementKey, element) = sortedDictionary.last { try serializeNext(element, forKey: elementKey) }
+        
+        if let (elementKey, element) = sortedDictionary.last {
+            try serializeNext(element, forKey: serializeNestedKey(elementKey, forKey: key))
+        }
     }
 }
 
--- a/Tests/OpenAPIRuntimeTests/URICoder/Encoding/Test_URIEncoder.swift
+++ b/Tests/OpenAPIRuntimeTests/URICoder/Encoding/Test_URIEncoder.swift
@@ -23,4 +23,12 @@ final class Test_URIEncoder: Test_Runtime {
         let encodedString = try encoder.encode(Foo(bar: "hello world"), forKey: "root")
         XCTAssertEqual(encodedString, "bar=hello+world")
     }
+    
+    func testNestedEncoding() throws {
+        struct Foo: Encodable { var bar: String }
+        let serializer = URISerializer(configuration: .deepObjectExplode)
+        let encoder = URIEncoder(serializer: serializer)
+        let encodedString = try encoder.encode(Foo(bar: "hello world"), forKey: "root")
+        XCTAssertEqual(encodedString, "root%5Bbar%5D=hello%20world")
+    }
 }
--- a/Tests/OpenAPIRuntimeTests/URICoder/Parsing/Test_URIParser.swift
+++ b/Tests/OpenAPIRuntimeTests/URICoder/Parsing/Test_URIParser.swift
@@ -17,7 +17,7 @@ import XCTest
 final class Test_URIParser: Test_Runtime {
 
     let testedVariants: [URICoderConfiguration] = [
-        .formExplode, .formUnexplode, .simpleExplode, .simpleUnexplode, .formDataExplode, .formDataUnexplode,
+        .formExplode, .formUnexplode, .simpleExplode, .simpleUnexplode, .formDataExplode, .formDataUnexplode, .deepObjectExplode
     ]
 
     func testParsing() throws {
@@ -29,7 +29,8 @@ final class Test_URIParser: Test_Runtime {
                     simpleExplode: .custom("", value: ["": [""]]),
                     simpleUnexplode: .custom("", value: ["": [""]]),
                     formDataExplode: "empty=",
-                    formDataUnexplode: "empty="
+                    formDataUnexplode: "empty=",
+                    deepObjectExplode: "empty="
                 ),
                 value: ["empty": [""]]
             ),
@@ -40,7 +41,8 @@ final class Test_URIParser: Test_Runtime {
                     simpleExplode: .custom("", value: ["": [""]]),
                     simpleUnexplode: .custom("", value: ["": [""]]),
                     formDataExplode: "",
-                    formDataUnexplode: ""
+                    formDataUnexplode: "",
+                    deepObjectExplode: ""
                 ),
                 value: [:]
             ),
@@ -51,7 +53,8 @@ final class Test_URIParser: Test_Runtime {
                     simpleExplode: .custom("fred", value: ["": ["fred"]]),
                     simpleUnexplode: .custom("fred", value: ["": ["fred"]]),
                     formDataExplode: "who=fred",
-                    formDataUnexplode: "who=fred"
+                    formDataUnexplode: "who=fred",
+                    deepObjectExplode: "who=fred"
                 ),
                 value: ["who": ["fred"]]
             ),
@@ -62,7 +65,8 @@ final class Test_URIParser: Test_Runtime {
                     simpleExplode: .custom("Hello%20World", value: ["": ["Hello World"]]),
                     simpleUnexplode: .custom("Hello%20World", value: ["": ["Hello World"]]),
                     formDataExplode: "hello=Hello+World",
-                    formDataUnexplode: "hello=Hello+World"
+                    formDataUnexplode: "hello=Hello+World",
+                    deepObjectExplode: "hello=Hello%20World"
                 ),
                 value: ["hello": ["Hello World"]]
             ),
@@ -73,7 +77,8 @@ final class Test_URIParser: Test_Runtime {
                     simpleExplode: .custom("red,green,blue", value: ["": ["red", "green", "blue"]]),
                     simpleUnexplode: .custom("red,green,blue", value: ["": ["red", "green", "blue"]]),
                     formDataExplode: "list=red&list=green&list=blue",
-                    formDataUnexplode: "list=red,green,blue"
+                    formDataUnexplode: "list=red,green,blue",
+                    deepObjectExplode: "list=red&list=green&list=blue"
                 ),
                 value: ["list": ["red", "green", "blue"]]
             ),
@@ -93,7 +98,8 @@ final class Test_URIParser: Test_Runtime {
                     formDataUnexplode: .custom(
                         "keys=comma,%2C,dot,.,semi,%3B",
                         value: ["keys": ["comma", ",", "dot", ".", "semi", ";"]]
-                    )
+                    ), 
+                    deepObjectExplode: "comma=%2C&dot=.&semi=%3B"
                 ),
                 value: ["semi": [";"], "dot": ["."], "comma": [","]]
             ),
@@ -133,6 +139,7 @@ extension Test_URIParser {
             static let simpleUnexplode: Self = .init(name: "simpleUnexplode", config: .simpleUnexplode)
             static let formDataExplode: Self = .init(name: "formDataExplode", config: .formDataExplode)
             static let formDataUnexplode: Self = .init(name: "formDataUnexplode", config: .formDataUnexplode)
+            static let deepObjectExplode: Self = .init(name: "deepObjectExplode", config: .deepObjectExplode)
         }
         struct Variants {
 
@@ -161,6 +168,7 @@ extension Test_URIParser {
             var simpleUnexplode: Input
             var formDataExplode: Input
             var formDataUnexplode: Input
+            var deepObjectExplode: Input
         }
         var variants: Variants
         var value: URIParsedNode
--- a/Tests/OpenAPIRuntimeTests/URICoder/Serialization/Test_URISerializer.swift
+++ b/Tests/OpenAPIRuntimeTests/URICoder/Serialization/Test_URISerializer.swift
@@ -31,7 +31,8 @@ final class Test_URISerializer: Test_Runtime {
                     simpleExplode: "",
                     simpleUnexplode: "",
                     formDataExplode: "empty=",
-                    formDataUnexplode: "empty="
+                    formDataUnexplode: "empty=",
+                    deepObjectExplode: "empty="
                 )
             ),
             makeCase(
@@ -43,7 +44,8 @@ final class Test_URISerializer: Test_Runtime {
                     simpleExplode: "fred",
                     simpleUnexplode: "fred",
                     formDataExplode: "who=fred",
-                    formDataUnexplode: "who=fred"
+                    formDataUnexplode: "who=fred",
+                    deepObjectExplode: "who=fred"
                 )
             ),
             makeCase(
@@ -55,7 +57,8 @@ final class Test_URISerializer: Test_Runtime {
                     simpleExplode: "1234",
                     simpleUnexplode: "1234",
                     formDataExplode: "x=1234",
-                    formDataUnexplode: "x=1234"
+                    formDataUnexplode: "x=1234",
+                    deepObjectExplode: "x=1234"
                 )
             ),
             makeCase(
@@ -67,7 +70,8 @@ final class Test_URISerializer: Test_Runtime {
                     simpleExplode: "12.34",
                     simpleUnexplode: "12.34",
                     formDataExplode: "x=12.34",
-                    formDataUnexplode: "x=12.34"
+                    formDataUnexplode: "x=12.34",
+                    deepObjectExplode: "x=12.34"
                 )
             ),
             makeCase(
@@ -79,7 +83,8 @@ final class Test_URISerializer: Test_Runtime {
                     simpleExplode: "true",
                     simpleUnexplode: "true",
                     formDataExplode: "enabled=true",
-                    formDataUnexplode: "enabled=true"
+                    formDataUnexplode: "enabled=true",
+                    deepObjectExplode: "enabled=true"
                 )
             ),
             makeCase(
@@ -91,7 +96,8 @@ final class Test_URISerializer: Test_Runtime {
                     simpleExplode: "Hello%20World",
                     simpleUnexplode: "Hello%20World",
                     formDataExplode: "hello=Hello+World",
-                    formDataUnexplode: "hello=Hello+World"
+                    formDataUnexplode: "hello=Hello+World",
+                    deepObjectExplode: "hello=Hello%20World"
                 )
             ),
             makeCase(
@@ -103,7 +109,8 @@ final class Test_URISerializer: Test_Runtime {
                     simpleExplode: "red,green,blue",
                     simpleUnexplode: "red,green,blue",
                     formDataExplode: "list=red&list=green&list=blue",
-                    formDataUnexplode: "list=red,green,blue"
+                    formDataUnexplode: "list=red,green,blue",
+                    deepObjectExplode: nil
                 )
             ),
             makeCase(
@@ -118,7 +125,8 @@ final class Test_URISerializer: Test_Runtime {
                     simpleExplode: "comma=%2C,dot=.,semi=%3B",
                     simpleUnexplode: "comma,%2C,dot,.,semi,%3B",
                     formDataExplode: "comma=%2C&dot=.&semi=%3B",
-                    formDataUnexplode: "keys=comma,%2C,dot,.,semi,%3B"
+                    formDataUnexplode: "keys=comma,%2C,dot,.,semi,%3B",
+                    deepObjectExplode: "keys%5Bcomma%5D=%2C&keys%5Bdot%5D=.&keys%5Bsemi%5D=%3B"
                 )
             ),
         ]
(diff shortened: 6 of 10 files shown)


Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
```diff
@@ -61,13 +61,15 @@ extension URIParser {
...
             case .simple: return ["": [""]]
+            case .deepObject: return [:]
             }
         }
         switch (configuration.style, configuration.explode) {
         case (.form, true): return try parseExplodedFormRoot()
         case (.form, false): return try parseUnexplodedFormRoot()
         case (.simple, true): return try parseExplodedSimpleRoot()
         case (.simple, false): return try parseUnexplodedSimpleRoot()
+        case (.deepObject, _): return try parseExplodedDeepObjectRoot()
```
simonjbeaumont, line 72: My reading of the OAS is that `deepObject` is only valid with `explode: true`[^0]. If that's the case, we should probably guard against that somehow.

[^0]: https://spec.openapis.org/oas/latest.html#style-examples

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
```diff
@@ -205,6 +207,50 @@ extension URIParser {
...
+                    // Hit the key/value separator, so a value will follow.
+                    let secondValue = data.parseUpToCharacterOrEnd(pairSeparator)
+                    
+                    key = nestedKey(from: firstValue)
+                    value = secondValue
+                case .foundSecondOrEnd:
+                    // No key/value separator, treat the string as the key.
+                    key = nestedKey(from: firstValue)
+                    value = .init()
+                }
```
simonjbeaumont, line 249: Can you give an example for this case? The spec is light on examples and IIUC doesn't have an example that _isn't_ the `object[key]=value` form.

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
```diff
@@ -326,4 +372,35 @@ extension String.SubSequence {
...
+    
+    
+    /// Accumulates characters from the `startingCharacter` character provided,
+    /// until the `endingCharacter` is reached. Moves the underlying startIndex.
+    /// - Parameters:
+    ///   - startingCharacter: A character to start with.
+    ///   - endingCharacter: A character to stop at.
+    /// - Returns: The accumulated substring.
+    fileprivate mutating func parseBetweenCharacters(startingCharacter: Character, endingCharacter: Character) -> Self {
+        let startIndex = startIndex
```
simonjbeaumont, line 384: This seems redundant?

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
```diff
@@ -326,4 +372,35 @@ extension String.SubSequence {
...
+        guard startIndex != endIndex else { return .init() }
+        guard let startingCharacterIndex = firstIndex(of: startingCharacter) else { return self }
+        var currentIndex = startingCharacterIndex
+        
+//        var x = self.firstIndex(of: startingCharacter)
```
simonjbeaumont, line 389: Presumably this can now be removed?

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
```diff
@@ -326,4 +372,35 @@ extension String.SubSequence {
...
     }
+    
+    
+    /// Accumulates characters from the `startingCharacter` character provided,
+    /// until the `endingCharacter` is reached. Moves the underlying startIndex.
+    /// - Parameters:
+    ///   - startingCharacter: A character to start with.
+    ///   - endingCharacter: A character to stop at.
+    /// - Returns: The accumulated substring.
+    fileprivate mutating func parseBetweenCharacters(startingCharacter: Character, endingCharacter: Character) -> Self {
```
simonjbeaumont, line 383: This function seems to overlap significantly with `parseUpToCharacterOrEnd(_:)`, the only difference is that we move the index forward to the first occurrence of `startingCharacter` first. Do we need the duplication, or can we implement one in terms of the other?

Tests/OpenAPIRuntimeTests/URICoder/Test_URICodingRoundtrip.swift
```diff
@@ -387,16 +407,20 @@ final class Test_URICodingRoundtrip: Test_Runtime {
...
+//            configuration: .formDataUnexplode,
+//            variant: variants.formDataUnexplode
+//        )
         try testVariant(
-            name: "formDataUnexplode",
-            configuration: .formDataUnexplode,
-            variant: variants.formDataUnexplode
+            name: "deepObjectExplode",
+            configuration: .deepObjectExplode,
+            variant: variants.deepObjectExplode
```
simonjbeaumont, line 423: This appears to have disabled a bunch of tests.

simonjbeaumont (changes requested):
First off, thanks @kstefanou52 for taking a stab at this. I've taken a brief look over the code and left some comments. I imagine @czechboy0 will have some to add too.

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
kstefanou52: Yeap you are right, that's the case. Should I throw an error instead of ignoring the explode value?

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
kstefanou52: Yeap, I accidentally committed it.

Tests/OpenAPIRuntimeTests/URICoder/Test_URICodingRoundtrip.swift
kstefanou52: Yeap, I accidentally committed it.

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
kstefanou52: Absolutely, I forgot to delete it.

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
kstefanou52: You mean keeping just one of the functions and add an extra param in order to determine its functionality?

Sources/OpenAPIRuntime/Conversion/ParameterStyles.swift
```diff
@@ -26,6 +26,11 @@
     ///
     /// Details: https://datatracker.ietf.org/doc/html/rfc6570#section-3.2.2
     case simple
+    
+    /// The deepObject style.
+    ///
+    /// Details: Should update datatracker
```
czechboy0, line 32: Can you put a link to the RFC please?

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
czechboy0: Yes please
```suggestion
        case (.deepObject, true): return try parseExplodedDeepObjectRoot()
        case (.deepObject, false): // throw an error here
```

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
czechboy0: Yes let's be strict here and throw an error if the string seems malformed.

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
czechboy0: What'd be the extra param needed? I also agree we should be able to use the existing function.

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
```diff
@@ -205,6 +207,50 @@ extension URIParser {
...
+                    first: keyValueSeparator,
+                    second: pairSeparator
+                )
+                let key: Raw
+                let value: Raw
+                switch firstResult {
+                case .foundFirst:
+                    // Hit the key/value separator, so a value will follow.
+                    let secondValue = data.parseUpToCharacterOrEnd(pairSeparator)
+
```
czechboy0, line 242:
```suggestion
```

Sources/OpenAPIRuntime/URICoder/Serialization/URISerializer.swift
```diff
@@ -180,6 +181,9 @@ extension URISerializer {
         case (.simple, _):
             keyAndValueSeparator = nil
             pairSeparator = ","
+        case (.deepObject, _):
```
czechboy0, line 184: By my reading of the OpenAPI 3.1 specification, deepObject _must_ have a JSON object at the top, so an array would not be supported? https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#style-examples

Sources/OpenAPIRuntime/URICoder/Serialization/URISerializer.swift
```diff
@@ -228,8 +232,15 @@ extension URISerializer {
         case (.simple, false):
             keyAndValueSeparator = ","
             pairSeparator = ","
+        case (.deepObject, _):
```
czechboy0, line 235: I think we should again validate that explode is true, and throw an error on false.

Sources/OpenAPIRuntime/URICoder/Serialization/URISerializer.swift
```diff
@@ -228,8 +232,15 @@ extension URISerializer {
...
+            keyAndValueSeparator = "="
+            pairSeparator = "&"
         }
 
+        func serializeNestedKey(_ elementKey: String, forKey rootKey: String) -> String {
+            guard case .deepObject = configuration.style else { return elementKey }
+            return rootKey + "[" + elementKey + "]"
```
czechboy0, line 242: Can you check that we end up percent-escaping the `[` and `]` characters? Technically they should be escaped.

However, when parsing, we should be flexible and support both escaped and unescaped.

Tests/OpenAPIRuntimeTests/URICoder/Encoding/Test_URIEncoder.swift
```diff
@@ -23,4 +23,12 @@ final class Test_URIEncoder: Test_Runtime {
         let encodedString = try encoder.encode(Foo(bar: "hello world"), forKey: "root")
         XCTAssertEqual(encodedString, "bar=hello+world")
     }
+    
+    func testNestedEncoding() throws {
+        struct Foo: Encodable { var bar: String }
+        let serializer = URISerializer(configuration: .deepObjectExplode)
+        let encoder = URIEncoder(serializer: serializer)
+        let encodedString = try encoder.encode(Foo(bar: "hello world"), forKey: "root")
+        XCTAssertEqual(encodedString, "root%5Bbar%5D=hello+world")
```
czechboy0, line 32: Are we sure that deepObject uses `+` instead of `%20` for encoding spaces? Can you find it in the RFC?

Tests/OpenAPIRuntimeTests/URICoder/Serialization/Test_URISerializer.swift
```diff
@@ -103,7 +109,8 @@ final class Test_URISerializer: Test_Runtime {
                     simpleExplode: "red,green,blue",
                     simpleUnexplode: "red,green,blue",
                     formDataExplode: "list=red&list=green&list=blue",
-                    formDataUnexplode: "list=red,green,blue"
+                    formDataUnexplode: "list=red,green,blue",
+                    deepObjectExplode: "list=red&list=green&list=blue"
```
czechboy0, line 113: Let's figure out what we do with arrays, I think either we don't support them, or they should be something like `list[]=red&...` or `list[0]=red&...`

Tests/OpenAPIRuntimeTests/URICoder/Serialization/Test_URISerializer.swift
```diff
@@ -118,7 +125,8 @@ final class Test_URISerializer: Test_Runtime {
                     simpleExplode: "comma=%2C,dot=.,semi=%3B",
                     simpleUnexplode: "comma,%2C,dot,.,semi,%3B",
                     formDataExplode: "comma=%2C&dot=.&semi=%3B",
-                    formDataUnexplode: "keys=comma,%2C,dot,.,semi,%3B"
+                    formDataUnexplode: "keys=comma,%2C,dot,.,semi,%3B",
+                    deepObjectExplode: "keys%5Bcomma%5D=%2C&keys%5Bdot%5D=.&keys%5Bsemi%5D=%3B"
```
czechboy0, line 129: Ah cool, this answers my question about escaping. Can you add another test case that verifies that we successfully parse strings with the `[` and `]` unescaped?

czechboy0 (changes requested):
Thanks for kicking off this work, @kstefanou52! A few comments, but overall this is the right approach.

Tests/OpenAPIRuntimeTests/URICoder/Encoding/Test_URIEncoder.swift
kstefanou52: Nice catch! Unfortunately RFC doesn't refer anything regarding deepObject. My reading of OpenAPISpec[[1]](https://spec.openapis.org/oas/v3.1.0.html#style-values) is that deepObject uses form parameters configuration, so I think we should use `%20` encoding for spaces.

## 
 1. https://spec.openapis.org/oas/v3.1.0.html#style-values

Sources/OpenAPIRuntime/Conversion/ParameterStyles.swift
kstefanou52: Unfortunately RFC doesn't refer anything regarding deepObject. My reading of OpenAPISpec[[1]](https://spec.openapis.org/oas/v3.1.0.html#style-values) is that deepObject uses form parameters configuration. 
Should we just refer to form style here too? What do you think?

## 
 1. https://spec.openapis.org/oas/v3.1.0.html#style-values

Tests/OpenAPIRuntimeTests/URICoder/Serialization/Test_URISerializer.swift
czechboy0: Not sure, I'd like to see what other generators are doing. Also, @simonjbeaumont, any thoughts?

Tests/OpenAPIRuntimeTests/URICoder/Encoding/Test_URIEncoder.swift
czechboy0: Yeah that makes sense, thanks for checking.

Sources/OpenAPIRuntime/Conversion/ParameterStyles.swift
czechboy0: Yeah link to the OpenAPI doc, to the closest section, that's good enoguh.

Tests/OpenAPIRuntimeTests/URICoder/Serialization/Test_URISerializer.swift
kstefanou52: I created a very simple example on [swagger](https://editor-next.swagger.io).

**Yaml file:**
```
---
openapi: 3.0.0
info:
  title: DeepObject Example
  version: TODO
servers:
- url: http://localhost:3000
  description: Development
paths:
  "/example":
    get:
      responses:
        '200':
          description: OK
      parameters:
        - in: query
          name: include
          schema:
            type: object
            properties:
              member:
                type: array
                example: [foo, bar]
          style: deepObject
          explode: true
  
```

**Generated request:**
```
http://localhost:3000/example?include[member]=foo&include[member]=bar
```

I don't know if that helps, it didn't helped me ..

simonjbeaumont: Unless we can find some stronger evidence for how it should be, I'd rather us not support it at all.

I took a quick look at `opai-codegen` and it seems they only test the object case (cf. array)[^1].

[^1]: https://github.com/deepmap/oapi-codegen/blob/dd082985a9b6e8f68472987f6c2c60fea5e59871/internal/test/parameters/parameters_test.go#L430-L436

simonjbeaumont: > ```
> http://localhost:3000/example?include[member]=foo&include[member]=bar
> ```

Oh, from the example above, it looks like arrays are serialized as repeated values for the same key. That's gonna complicate the logic some, I think. Be good to see if we can find any documentation that supports this.

czechboy0: Ok let's just not support non-object values when deepObject is used, and throw appropriate errors. We can always add it later once we get more clarity.

alephao:
Hello, looks like all the test cases except for the `dictionary` are wrong since the behavior of non-deep objects is undefined. Am I missing something?

Seems like undefined behavior from deepObjectExplode is falling back to formDataExplode, is that intentional?

czechboy0:
@alephao yes these were called out in the review, see the code comments and expand the resolved ones (they don't seem to actually be updated in code yet).

alephao:
@czechboy0 Oh, sorry, I got a bit lost in the comments and missed that one!

@kstefanou52 since this been stuck for a few weeks, I sent a PR to your fork addressing most of the requested changes (not sure if it's all there cuz I got a bit lost in the comments) https://github.com/kstefanou52/swift-openapi-runtime/pull/1

This is a summary of the changes:
- fix URISerializer for deepObject and its tests (following the spec and not handling undefined behavior)
- add deepObject tests to URIParser (there were no tests here)
- fix URICodingRoundtrip tests (I ended up ignoring the cases for deepObject when the behavior is undefined since roundtrip doesn't make much sense to me in those cases)
- remove parseBetweenCharacters function (I used parseUpToCharacterOrEnd twice instead)
- add preconditionFailure to code that should not be reachable (serializeArray with deepObject)

kstefanou52:
@czechboy0 @alephao Sorry I was away from keyboard the last 2 weeks due to vacations. I have fixed the majority of the comments and I'll push it at the of the day.

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
```diff
@@ -302,17 +349,55 @@ extension String.SubSequence {
...
     /// or the end is reached. Moves the underlying startIndex.
-    /// - Parameter character: A character to stop at.
+    /// - Parameters:
+    ///   - startingCharacter: A character to start with.
+    ///   - endingCharacter: A character to stop at.
+    ///   If not provided or not found then uses the current start index as a starting character.
     /// - Returns: The accumulated substring.
-    fileprivate mutating func parseUpToCharacterOrEnd(_ character: Character) -> Self {
-        let startIndex = startIndex
+    fileprivate mutating func parseUpToCharacterOrEnd(startingCharacter: Character? = nil, _ endingCharacter: Character) -> Self {
```
czechboy0, line 360: Can we take one more look here? Could you try to revert this part and go back to the original parseUpToCharacterOrEnd(_ character: Character) function, and explain why it can't work for you?

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
```diff
@@ -205,6 +213,45 @@ extension URIParser {
...
+    /// - Throws: An error if parsing fails.
+    private mutating func parseExplodedDeepObjectRoot() throws -> URIParsedNode {
+        try parseGenericRoot { data, appendPair in
+            let keyValueSeparator: Character = "="
+            let pairSeparator: Character = "&"
+            let nestedKeyStartingCharacter: Character = "["
+            let nestedKeyEndingCharacter: Character = "]"
+            
+            func nestedKey(from deepObjectKey: String.SubSequence) -> Raw {
+                var unescapedFirstValue = Substring(deepObjectKey.removingPercentEncoding ?? "")
```
czechboy0, line 229: Here you should verify that the first character is `[`, otherwise this is a malformed key. For the rest of the characters, you should be able to use `parseUpToCharacterOrEnd` as originally was.

Sources/OpenAPIRuntime/URICoder/Serialization/URISerializer.swift
czechboy0: This should throw an error - deepObject doesn't support arrays.

Sources/OpenAPIRuntime/URICoder/Serialization/URISerializer.swift
czechboy0: Thanks for adding the test verifying this.

Tests/OpenAPIRuntimeTests/URICoder/Parsing/Test_URIParser.swift
```diff
@@ -29,7 +29,8 @@ final class Test_URIParser: Test_Runtime {
                     simpleExplode: .custom("", value: ["": [""]]),
                     simpleUnexplode: .custom("", value: ["": [""]]),
                     formDataExplode: "empty=",
-                    formDataUnexplode: "empty="
+                    formDataUnexplode: "empty=",
+                    deepObjectExplode: "empty="
```
czechboy0, line 33: Shouldn't this have some square brackets?

Tests/OpenAPIRuntimeTests/URICoder/Parsing/Test_URIParser.swift
```diff
@@ -51,7 +53,8 @@ final class Test_URIParser: Test_Runtime {
                     simpleExplode: .custom("fred", value: ["": ["fred"]]),
                     simpleUnexplode: .custom("fred", value: ["": ["fred"]]),
                     formDataExplode: "who=fred",
-                    formDataUnexplode: "who=fred"
+                    formDataUnexplode: "who=fred",
+                    deepObjectExplode: "who=fred"
```
czechboy0, line 57: Shouldn't this have some square brackets?

Tests/OpenAPIRuntimeTests/URICoder/Parsing/Test_URIParser.swift
```diff
@@ -62,7 +65,8 @@ final class Test_URIParser: Test_Runtime {
                     simpleExplode: .custom("Hello%20World", value: ["": ["Hello World"]]),
                     simpleUnexplode: .custom("Hello%20World", value: ["": ["Hello World"]]),
                     formDataExplode: "hello=Hello+World",
-                    formDataUnexplode: "hello=Hello+World"
+                    formDataUnexplode: "hello=Hello+World",
+                    deepObjectExplode: "hello=Hello%20World"
```
czechboy0, line 69: Shouldn't this have some square brackets?

Tests/OpenAPIRuntimeTests/URICoder/Parsing/Test_URIParser.swift
```diff
@@ -73,7 +77,8 @@ final class Test_URIParser: Test_Runtime {
                     simpleExplode: .custom("red,green,blue", value: ["": ["red", "green", "blue"]]),
                     simpleUnexplode: .custom("red,green,blue", value: ["": ["red", "green", "blue"]]),
                     formDataExplode: "list=red&list=green&list=blue",
-                    formDataUnexplode: "list=red,green,blue"
+                    formDataUnexplode: "list=red,green,blue",
+                    deepObjectExplode: "list=red&list=green&list=blue"
```
czechboy0, line 81: Shouldn't this have some square brackets? Or be unsupported, as an array?

Tests/OpenAPIRuntimeTests/URICoder/Parsing/Test_URIParser.swift
```diff
@@ -93,7 +98,8 @@ final class Test_URIParser: Test_Runtime {
                     formDataUnexplode: .custom(
                         "keys=comma,%2C,dot,.,semi,%3B",
                         value: ["keys": ["comma", ",", "dot", ".", "semi", ";"]]
-                    )
+                    ), 
+                    deepObjectExplode: "comma=%2C&dot=.&semi=%3B"
```
czechboy0, line 102: Shouldn't this have some square brackets?

Tests/OpenAPIRuntimeTests/URICoder/Serialization/Test_URISerializer.swift
```diff
@@ -31,7 +31,8 @@ final class Test_URISerializer: Test_Runtime {
                     simpleExplode: "",
                     simpleUnexplode: "",
                     formDataExplode: "empty=",
-                    formDataUnexplode: "empty="
+                    formDataUnexplode: "empty=",
+                    deepObjectExplode: "empty="
```
czechboy0, line 35: Same in this whole file: Shouldn't this have some square brackets?

czechboy0 (changes requested):
Thanks for the updates, a few more comments

alephao:
Hi @kstefanou52, I'm sorry that I'm being annoying but I really want this change in the upstream and the PR seem to be stuck, I updated my PR to your fork with your new changes, it already addresses all the requested changes by @czechboy0, please have a look there or if you don't have time to keep working on this I'm happy to open another PR and I can put my focus on it and finish the work you've been doing.

Again, sorry to push, I just want this feature merged soon!

kstefanou52:
@alephao I'll push my changes until the end of day.

Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
kstefanou52: Certainly. I need a method that can extract the nested key "bar" from the key: foo[bar]. The original method doesn't allow me to retrieve this nested key. When I pass "[" as a parameter, I receive "foo", and when I pass "]" I get "foo[bar". Am I missing something here?

czechboy0: Right, that needs to happen in two steps:
1. Pass `[`, and get `foo` (the top level key)
2. Increment the iterator to the next character, because you know you are currently pointed at `[`.
3. Pass `]`, and get `bar` (the nested key)
4. Increment the iterator to move beyond the `]`.

kstefanou52: Ok now I got it. Much better this way. Thank you :)

Tests/OpenAPIRuntimeTests/URICoder/Parsing/Test_URIParser.swift
kstefanou52: You mean the correct output should be `[hello]=Hello%20World` ?

kstefanou52:
Hello @simonjbeaumont @czechboy0 ! I just pushed my changes. the only thing that I'm still missing are the comments you left about Test_URiParser with the square brackets. I think that the tests are correct because it tests the nested parsing. Am I missing something ? thank you so much for the guidance and sorry for the delay. cheers!

alephao:
@kstefanou52 thanks for the updates! I'm going to try to explain @czechboy0's points since I do agree with the changes he has requested.

Following the matrix [here](https://spec.openapis.org/oas/latest.html#style-examples), the spec doesn't define the behavior of "empty", "string" or "array" unless the document is using a spec extension.

My interpretation of it is that primitives like `"foo=bar"`, `"foo=1"`, should throw an error or at least be ignored and the runtime should not attempt to parse them, here is my reasoning:

**Imagine the following concrete example:**

I have an endpoint to update user's first and last name, I like to model my sever using the deepObject style, so the decoders on my server work on inputs like `user[first_name]=string&user[last_name]`. My server will not handle `first_name=string&last_name=string` and I specify this to the users of my api by writing on my openapi spec that the endpoint accepts only deepObject structures. So if you try to send anything other than a deepObject, my server will return an error.

Now imagine a mocked server generated by this library, if it accepts `first_name=foo&last_name=bar`, it is falling to correctly mock the server, since I specifically told in my openapi spec that I only accept deepObject structures. Once I finish development and plug my code to the real server, I'll have a surprise: my code doesn't work on the live server, only on the mock server, because the mock server successfully parses inputs that I specifically told I do not accept.

Now what that means in terms of code that needs to be changed here is that all the URIParser's tests should stop handling the "primitive=value" case, there should be an assertion that an empty dictionary was generated (meaning that the incorrect input was ignored) or it should assert that an error was thrown. 

When looking at the empty test case of the `style = simple`, I'm deducing that we should do the first option: just ignore the invalid input and return an empty dictionary.

Please let me know if this makes any sense or if I'm completely wrong

kstefanou52:
@alephao if I'm not wrong @czechboy0 requested to oarse using square braces and not ignore them. But to be honest I'm not completely sure. Let's wait for his response.

alephao (commented):
@kstefanou52 about the brackets check the suggestions below, there is also a request to create test cases for non-escaped brackets

czechboy0:
Yeah @alephao is exactly right here - unsupported types of objects need to throw an error, but should not parse as if they were simple/form.

Tests/OpenAPIRuntimeTests/URICoder/Parsing/Test_URIParser.swift
czechboy0: That's good - then this test should reflect that. Either by catching an expected error or expecting an empty parsed dictionary.

Tests/OpenAPIRuntimeTests/URICoder/Serialization/Test_URISerializer.swift
```diff
@@ -140,6 +148,9 @@ final class Test_URISerializer: Test_Runtime {
             try testVariant(.simpleUnexplode, testCase.variants.simpleUnexplode)
             try testVariant(.formDataExplode, testCase.variants.formDataExplode)
             try testVariant(.formDataUnexplode, testCase.variants.formDataUnexplode)
+            if let deepObjectExplode = testCase.variants.deepObjectExplode {
```
czechboy0, line 151: Why is this if statement here instead of running the variant unconditionally like the previous few?

czechboy0 (changes requested):
The code looks okay now, just the tests need updating based on @alephao and my comments. 🙏

Tests/OpenAPIRuntimeTests/URICoder/Serialization/Test_URISerializer.swift
kstefanou52: Yeah, I don't like it either. My problem here is the scenario that we are testing arrays serialisation. 
I managed to add an `expectedError` on the makeCase in order to handle errors too.

Tests/OpenAPIRuntimeTests/URICoder/Parsing/Test_URIParser.swift
kstefanou52: Completed. I have added support for testing the errors thrown as well. However, I'm unsure about whether it's correct that I made `ParsingError` and `SerializationError` public. I would appreciate your feedback on this.

Tests/OpenAPIRuntimeTests/URICoder/Serialization/Test_URISerializer.swift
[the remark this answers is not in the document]
czechboy0: This applies to the whole file and the next one.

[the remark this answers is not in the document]
alephao: > I think primitive values should also throw an error, as only dictionaries are supported in deepObjects, isn't it?

I believe that's the case too! If a param `obj` has the style `deepObject/explode`, then in the query `obj[foo]=bar&obj=foo`, `obj=foo` doesn't make sense and shouldn't be allowed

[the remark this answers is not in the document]
kstefanou52: I concur 👍

kstefanou52:
Hello @czechboy0 and @alephao, could you please take a look at `Test_URICodingRoundtrip`? I'm not entirely confident that all the test scenarios are correct, especially those involving a struct with a custom Codable implementation that forwards, and handling `AnyOf` cases. Thanks!

alephao (commented):
Hi @kstefanou52, let me know if I'm wrong, but I believe that the point of these roundtrip tests is to check that a successfully encoded value can also be decoded, all the tests expecting errors won't encode nor decode anything, so they end up missing the point.

I think would be nice if we could do something like this:

```swift
// Note: This code won't work, it's just an idea
deepObjectExplode: .custom(
   "obj[who]=fred",
   value: ["who": "fred"]
)
```

But I pulled your latest changes and spent a bit of time hammering a way to test the deepObject in the current setup (with the `_test` function), but I wasn't happy with anything I tried, code was getting spaghetti and hard to reason about.

So I tried writing a separate function to test deepObjects which ended up working well. It's almost the same as the generic code in `_test`.

```swift
// func definition
func _testDeepObject<T: Codable & Equatable>(key: String, value: T, encodedString string: String, file: StaticString = #file, line: UInt = #line) throws {
    let encoder = URIEncoder(configuration: .deepObjectExplode)
    let encodedString = try encoder.encode(value, forKey: key)
    XCTAssertEqual(encodedString, string, "Variant: deepObjectExplode", file: file, line: line)
    let decoder = URIDecoder(configuration: .deepObjectExplode)
    let decodedValue = try decoder.decode(T.self, forKey: key, from: encodedString[...])
    XCTAssertEqual(decodedValue, value, "Variant: deepObjectExplode", file: file, line: line)
}

// usage example
try _testDeepObject(key: "obj", value: ["root": "Hello World!"], encodedString: "obj%5Broot%5D=Hello%20World%21")
```

I'm not sure if it's the best approach, but it does work and it fulfills what I believe to be the goal of these tests

kstefanou52:
Thank you @alephao for reviewing it. I completely agree with your comment. However, since we are only supporting encoding dictionaries and not primitive values, I thought it would be appropriate to throw the necessary error and test this behavior. Let's wait for @czechboy0 to confirm how we should proceed.

kstefanou52:
@czechboy0, are there any further changes needed? I apologize for being a bit lost with all these conversations. Thank you!

czechboy0 (approved):
Ok looks good to me! Let's run the CI and get that to green now.

czechboy0:

czechboy0:
The API breakage failure is expected:
> 💔 API breakage: enumelement ParameterStyle.deepObject has been added as a new enum case

It's in the SPI that only generated code uses, so it's okay to allow this "breakage" (it's not really, it was designed to be extended).

czechboy0:
Ok soundness failed, please run this locally and check in the changes:

> ./scripts/run-swift-format.sh --fix

kstefanou52:

czechboy0:

kstefanou52:
@czechboy0, I think the soundness check must be green now.

czechboy0:

PR #102, opened 2024-04-02
Add support for application/xml body

ugocottin:
### Motivation

See [apple/swift-openapi-generator#556](https://github.com/apple/swift-openapi-generator/issues/556) for more.

### Modifications

Add converter methods for encoding and decoding XML request and response body.
Add `CustomCoder` protocol, allows to use other `Encoder` and `Decoder` for other `content-type` body.

User can define custom coder, and assign a custom coder to a specific content-type within `Configuration.customCoders` dictionary.

### Result

It's now possible to define custom encoder and decoder for supported content-type.

### Test Plan

Added converter methods are tested with a mock custom coder for `application/xml` content-type. To avoid adding a dependency to a XMLCoder like [CoreOffice/XMLCoder](https://github.com/CoreOffice/XMLCoder), mock custom coder uses JSONEncoder and JSONDecoder.

Encoding and decoding to XML are out of scope of the tests, because encoding and decoding logic must be provided by user through custom coder implementation.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/Configuration.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Configuration.swift
@@ -123,7 +123,8 @@ public struct Configuration: Sendable {
     /// The generator to use when creating mutlipart bodies.
     public var multipartBoundaryGenerator: any MultipartBoundaryGenerator
 
-    public var customCoders: [String: any CustomCoder]
+    /// Custom XML coder for encoding and decoding xml bodies.
+    public var xmlCoder: (any CustomCoder)?
     
     /// Creates a new configuration with the specified values.
     ///
@@ -131,19 +132,15 @@ public struct Configuration: Sendable {
     ///   - dateTranscoder: The transcoder to use when converting between date
     ///   and string values.
     ///   - multipartBoundaryGenerator: The generator to use when creating mutlipart bodies.
-    ///   - customCoders: Array of custom coder to use for encoding and decoding unsupported content types.
+    ///   - xmlCoder: Custom XML coder for encoding and decoding xml bodies.
     public init(
         dateTranscoder: any DateTranscoder = .iso8601,
         multipartBoundaryGenerator: any MultipartBoundaryGenerator = .random,
-        customCoders: [String: any CustomCoder] = [:]
+        xmlCoder: (any CustomCoder)? = nil
     ) {
         self.dateTranscoder = dateTranscoder
         self.multipartBoundaryGenerator = multipartBoundaryGenerator
-        self.customCoders = customCoders
+        self.xmlCoder = xmlCoder
     }
-
-    public func customCoder(for contentType: String) -> (any CustomCoder)? {
-        self.customCoders[contentType]
-    }
-
+    
 }
--- a/Sources/OpenAPIRuntime/Conversion/Configuration.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Configuration.swift
@@ -128,6 +128,7 @@ public struct Configuration: Sendable {
 
     /// Custom XML coder for encoding and decoding xml bodies.
     public var xmlCoder: (any CustomCoder)?
+
     /// Creates a new configuration with the specified values.
     ///
     /// - Parameters:
--- a/Sources/OpenAPIRuntime/Conversion/Configuration.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Configuration.swift
@@ -125,7 +125,6 @@ public struct Configuration: Sendable {
 
     /// Custom XML coder for encoding and decoding xml bodies.
     public var xmlCoder: (any CustomCoder)?
-    
     /// Creates a new configuration with the specified values.
     ///
     /// - Parameters:
@@ -142,5 +141,4 @@ public struct Configuration: Sendable {
         self.multipartBoundaryGenerator = multipartBoundaryGenerator
         self.xmlCoder = xmlCoder
     }
-    
 }
--- a/Sources/OpenAPIRuntime/Conversion/Configuration.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Configuration.swift
@@ -134,7 +134,7 @@ public struct Configuration: Sendable {
     ///   - dateTranscoder: The transcoder to use when converting between date
     ///   and string values.
     ///   - multipartBoundaryGenerator: The generator to use when creating mutlipart bodies.
-    ///   - xmlCoder: Custom XML coder for encoding and decoding xml bodies.
+    ///   - xmlCoder: Custom XML coder for encoding and decoding xml bodies. Only required when using XML body payloads.
     public init(
         dateTranscoder: any DateTranscoder = .iso8601,
         multipartBoundaryGenerator: any MultipartBoundaryGenerator = .random,
--- a/Sources/OpenAPIRuntime/Conversion/Configuration.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Configuration.swift
@@ -101,15 +101,18 @@ public protocol CustomCoder: Sendable {
 
     /// Encodes the given value and returns its custom encoded representation.
     ///
-    /// - parameter value: The value to encode.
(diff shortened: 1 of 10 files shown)


Sources/OpenAPIRuntime/Conversion/CurrencyExtensions.swift
```diff
@@ -144,6 +144,32 @@ extension Converter {
         return HTTPBody(data)
     }
 
+    /// Returns a value decoded from a XML body.
+    /// - Parameter body: The body containing the raw XML bytes.
+    /// - Returns: A decoded value.
+    /// - Throws: An error if decoding from the body fails.
+    /// - Throws: An error if no custom coder is present for XML coding.
+    func convertXMLToBodyCodable<T: Decodable>(_ body: HTTPBody) async throws -> T {
+        guard let coder = configuration.customCoder(for: "application/xml") else {
```
czechboy0, line 153: Please add a static let for xml on the `OpenAPIMIMEType` type, and use that here, so that we don't need to copy the raw MIME type for XML 4 times in this file.

Sources/OpenAPIRuntime/Errors/RuntimeError.swift
```diff
@@ -88,6 +89,7 @@ internal enum RuntimeError: Error, CustomStringConvertible, LocalizedError, Pret
         case .invalidBase64String(let string):
             return "Invalid base64-encoded string (first 128 bytes): '\(string.prefix(128))'"
         case .failedToDecodeStringConvertibleValue(let string): return "Failed to decode a value of type '\(string)'."
+        case .missingCoderForCustomContentType(let contentType): return "Missing custom coder for content type '\(contentType)'"
```
czechboy0, line 92:
```suggestion
        case .missingCoderForCustomContentType(let contentType): return "Missing custom coder for content type '\(contentType)'."
```

Tests/OpenAPIRuntimeTests/Conversion/Test_Converter+Client.swift
```diff
@@ -120,6 +120,30 @@ final class Test_ClientConverterExtensions: Test_Runtime {
         try await XCTAssertEqualStringifiedData(body, testStructPrettyString)
         XCTAssertEqual(headerFields, [.contentType: "application/json", .contentLength: "23"])
     }
+    
+    //    | client | set | request body | XML | optional | setOptionalRequestBodyAsXML |
```
czechboy0, line 124: Reminder: Please adapt the `Converting-between-data-and-Swift-types.md` docs page with the addition of XML in your swift-openapi-generator PR, when it comes to it.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
```diff
@@ -105,17 +123,27 @@ public struct Configuration: Sendable {
     /// The generator to use when creating mutlipart bodies.
     public var multipartBoundaryGenerator: any MultipartBoundaryGenerator
 
+    public var customCoders: [String: any CustomCoder]
```
czechboy0, line 126: I'm a bit worried about this being too generic and folks assuming it'll override the JSON coders as well (which it won't).

Could you change this to `public var xmlCoder: (any CustomCoder)?` instead? I think it's better to be explicit here.

czechboy0 (changes requested):
Looks really good! A few initial comments, mostly minor suggestions.

czechboy0:

Sources/OpenAPIRuntime/Conversion/Configuration.swift
ugocottin: Should we also get rid of `Configuration.customCoder(for:)` method, and use property `xmlCoder` in `Converter.convertXMLToBodyCodable(_:)` and `Converter.convertBodyCodableToXML(_:)` ?

czechboy0: Yup exactly. Keep the public protocol `CustomCoder`, but only allow customizing the XML one on the Configuration struct.

ugocottin:
@czechboy0 I've made the suggested changes, the PR on swift-openapi-generator will follow soon, thanks

Sources/OpenAPIRuntime/Conversion/Configuration.swift
```diff
@@ -96,6 +96,24 @@ extension JSONDecoder.DateDecodingStrategy {
...
 }
 
+/// A type that allows custom content type encoding and decoding.
+public protocol CustomCoder: Sendable {
+
+    /// Encodes the given value and returns its custom encoded representation.
+    ///
+    /// - parameter value: The value to encode.
+    /// - returns: A new `Data` value containing the custom encoded data.
+    func customEncode<T: Encodable>(_ value: T) throws -> Data
```
czechboy0, line 106: Please run the `./scripts/soundness.sh` script and check it doesn't produce any warnings, I think these comments aren't entirely consistent with the rest.

Sources/OpenAPIRuntime/Conversion/Configuration.swift
```diff
@@ -105,17 +123,24 @@ public struct Configuration: Sendable {
...
+    /// Custom XML coder for encoding and decoding xml bodies.
+    public var xmlCoder: (any CustomCoder)?
+    
     /// Creates a new configuration with the specified values.
     ///
     /// - Parameters:
     ///   - dateTranscoder: The transcoder to use when converting between date
     ///   and string values.
     ///   - multipartBoundaryGenerator: The generator to use when creating mutlipart bodies.
+    ///   - xmlCoder: Custom XML coder for encoding and decoding xml bodies.
```
czechboy0, line 134:
```suggestion
    ///   - xmlCoder: Custom XML coder for encoding and decoding xml bodies. Only required when using XML body payloads.
```

Sources/OpenAPIRuntime/Conversion/Configuration.swift
```diff
@@ -105,17 +123,24 @@ public struct Configuration: Sendable {
...
     public init(
         dateTranscoder: any DateTranscoder = .iso8601,
-        multipartBoundaryGenerator: any MultipartBoundaryGenerator = .random
+        multipartBoundaryGenerator: any MultipartBoundaryGenerator = .random,
+        xmlCoder: (any CustomCoder)? = nil
```
czechboy0, line 138: You'll also need to add a deprecated version of the 2-param initializer into `Deprecated.swift`, as adding a parameter is technically an API-breaking change, so we also need the previous one, and it'll just call back into this one. Something like:

```swift
extension Configuration {
    /// Creates a new payload.
    @available(*, deprecated, renamed: "init(dateTranscoder:multipartBoundaryGenerator:xmlCoder:)") @_disfavoredOverload
    public init(
        multipartBoundaryGenerator: any MultipartBoundaryGenerator = .random
    ) {
        self.init(
            dateTranscoder: dateTranscoder,
            multipartBoundaryGenerator: multipartBoundaryGenerator,
            xmlCoder: nil
        )
    }
}
```

Sources/OpenAPIRuntime/Conversion/Configuration.swift
ugocottin: Done

czechboy0:

ugocottin:
Jenkins CI reports that `OpenAPIMIMEType` struct is not `Sendable`, and must be to be use in `RuntimeError.missingCoderForCustomContentType(contentType:)`. 

`OpenAPIMIMEType` seems thread safe to me. Should I make `OpenAPIMIMEType` conform to `Sendable`, or use String type in `RuntimeError.missingCoderForCustomContentType(contentType:)` ?

czechboy0:
Just convert it to string when constructing the error.

ugocottin:

czechboy0:

czechboy0 (approved):

czechboy0:

czechboy0:
Looks great, @ugocottin - will merge after CI passes. Feel free to move onto the generator PR. We'll do a release of the runtime early next week, at which point your generator PR should start passing on CI.

PR #104, opened 2024-04-08
Improve error message when encountering unexpected content-type headers

simonjbeaumont:
### Motivation

When the server response has a content-type header that does not conform to the OpenAPI document, we currently throw an error. However, this error presents itself in a very confusing manner: it prints `Unexpected Content-Type header: application/json`, where `application/json is the _expected_ content-type. At best, this is ambiguous and potentially misleading.

### Modifications

- Extend `case RuntimeError.unexpectedContentTypeHeader` with _both_ the expected and received content-type associated values.
- Update the printed description to include both the expected and received content-type header values.

### Result

When an unexpected content-type is received, the error message is clearer.

### Test Plan

Updated the existing test that expects an error to check the error and that the error values are provided in the correct order.


czechboy0 (approved):
Nice quality of life improvement! 👏

PR #107, opened 2024-06-20
Fixing Swift 6 Compatibility on Linux Issues

leogdion:
### Motivation

Fixing Issues with Building in Swift 6 for Linux

### Modifications

* Make Static Variables nonisolated safe
* Fix Issues with Upstream Iterator

### Result

Compatability with Swift 6 on Linux

### Test Plan

TBD

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 6.0
+// swift-tools-version: 5.9
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the SwiftOpenAPIGenerator open source project
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.9
+// swift-tools-version: 6.0
 //===----------------------------------------------------------------------===//
 //
 // This source file is part of the SwiftOpenAPIGenerator open source project
--- a/Sources/OpenAPIRuntime/Conversion/URLExtensions.swift
+++ b/Sources/OpenAPIRuntime/Conversion/URLExtensions.swift
@@ -14,13 +14,21 @@
 import Foundation
 
 extension URL {
+  private static func defaultOpenAPIServer () -> Self {
+    guard let url = URL(string: "/") else { fatalError("Failed to create an URL with the string '/'.") }
+    return url
+  }
+#if swift(>=6.0)
     /// Returns the default server URL of "/".
     ///
     /// Specification: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields
-    public nonisolated(unsafe) static let defaultOpenAPIServerURL: Self = {
-        guard let url = URL(string: "/") else { fatalError("Failed to create an URL with the string '/'.") }
-        return url
-    }()
+  public nonisolated(unsafe) static let defaultOpenAPIServerURL: Self = defaultOpenAPIServer()
+  #else
+  /// Returns the default server URL of "/".
+  ///
+  /// Specification: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields
+public static let defaultOpenAPIServerURL: Self = defaultOpenAPIServer()
+  #endif
 
     /// Returns a validated server URL, or throws an error.
     /// - Parameter string: A URL string.
--- a/Sources/OpenAPIRuntime/Conversion/URLExtensions.swift
+++ b/Sources/OpenAPIRuntime/Conversion/URLExtensions.swift
@@ -18,23 +18,24 @@ extension URL {
     guard let url = URL(string: "/") else { fatalError("Failed to create an URL with the string '/'.") }
     return url
   }
-#if swift(>=6.0)
+  
+  #if swift(>=6.0)
     /// Returns the default server URL of "/".
     ///
     /// Specification: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields
-  public nonisolated(unsafe) static let defaultOpenAPIServerURL: Self = defaultOpenAPIServer()
+    public nonisolated(unsafe) static let defaultOpenAPIServerURL: Self = defaultOpenAPIServer()
   #else
-  /// Returns the default server URL of "/".
-  ///
-  /// Specification: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields
-public static let defaultOpenAPIServerURL: Self = defaultOpenAPIServer()
+    /// Returns the default server URL of "/".
+    ///
+    /// Specification: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields
+    public static let defaultOpenAPIServerURL: Self = defaultOpenAPIServer()
   #endif
-
-    /// Returns a validated server URL, or throws an error.
-    /// - Parameter string: A URL string.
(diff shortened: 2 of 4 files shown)


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.9
+// swift-tools-version: 6.0
```
czechboy0, line 1: This would break existing users on pre-6.0 toolchains, so we can't do it this way.

sebsto: I would propose to have two `Package.swift` files, one generic and one specific for Swift 6.0 (`Package@swift-6.0.swift`)

sebsto:
Can we embed the `nonisolated(unsafe)` section with `#if swift(>=6.0)` to allow to continue to use Swift < 6.0 toolchains ?

Package.swift
czechboy0: We likely won't adopt any features that require Swift 6 while we still support Swift 5, so the first goal here would be to fix as many warnings as possible that come up with `-Xswiftc -strict-concurrency=complete`.

Sources/OpenAPIRuntime/Conversion/URLExtensions.swift
```diff
@@ -17,7 +17,7 @@ extension URL {
     /// Returns the default server URL of "/".
     ///
     /// Specification: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields
-    public static let defaultOpenAPIServerURL: Self = {
+    public nonisolated(unsafe) static let defaultOpenAPIServerURL: Self = {
```
czechboy0, line 20: This would be better fixed by Foundation.URL being Sendable on Linux (it already is on Darwin).

Sources/OpenAPIRuntime/URICoder/Serialization/URISerializer.swift
```diff
@@ -50,16 +50,16 @@ extension CharacterSet {
 
     /// A character set of unreserved symbols only from RFC 6570 (excludes
     /// alphanumeric characters).
-    fileprivate static let unreservedSymbols: CharacterSet = .init(charactersIn: "-._~")
+    fileprivate nonisolated(unsafe) static let unreservedSymbols: CharacterSet = .init(charactersIn: "-._~")
```
czechboy0, line 53: This would be better fixed by Foundation.CharacterSet being Sendable on Linux (it already is on Darwin).

Sources/OpenAPIRuntime/Multipart/MultipartFramesToRawPartsSequence.swift
```diff
@@ -342,8 +342,13 @@ extension MultipartFramesToRawPartsSequence {
             switch stateMachine.nextFromPartSequence() {
             case .returnNil: return nil
             case .fetchFrame:
+              let frame : Upstream.AsyncIterator.Element?
```
czechboy0, line 345: Could you break out these async sequence iteration fixes into a separate PR? These I think we'll be able to land soon, but the other ones (CharacterSet and URL), I'd like us to fix in swift-corelibs-foundation instead.

The shape of this fix is correct, but I wonder if we can make the `#if` more selective to the stdlib having `next(isolation:)`, rather than to switch by platform?

czechboy0: Let's use `#if compiler(>=6.0)`.

leogdion:
@czechboy0  I'm not sure in the case `URL` or `CharacterSet` we can simply just apply `Sendable` to these types. 

[`URL` contains a reference to `NSURL` and `URLBox` which is not `Sendable`](https://github.com/apple/swift-corelibs-foundation/blob/ca3669eb9ac282c649e71824d9357dbe140c8251/Sources/Foundation/URL.swift#L438) and I'm not sure would be easy to apply. 
[Same with `CharacterSet` and `NSCharacterSet`.](https://github.com/apple/swift-corelibs-foundation/blob/ca3669eb9ac282c649e71824d9357dbe140c8251/Darwin/Foundation-swiftoverlay/CharacterSet.swift#L369)

Let me if I'm missing something or looking at the wrong thing.

Sources/OpenAPIRuntime/Multipart/MultipartFramesToRawPartsSequence.swift
leogdion: Can we specify `Linux` and `compiler(>=6.0)`? There's no issue compiling it in Xcode as is.

Sources/OpenAPIRuntime/Conversion/URLExtensions.swift
leogdion: For reference: https://github.com/apple/swift-openapi-runtime/pull/107#issuecomment-2180542772

Sources/OpenAPIRuntime/URICoder/Serialization/URISerializer.swift
leogdion: For reference: https://github.com/apple/swift-openapi-runtime/pull/107#issuecomment-2180542772

Sources/OpenAPIRuntime/Multipart/MultipartFramesToRawPartsSequence.swift
czechboy0: It's because complete concurrency checking isn't enabled by default, but if you turn it on, you'll see it on Darwin as well. No need to limit this to Linux, this is not a platform-specific issue, it's a Swift version-specific fix.

MahdiBM:
corelibs-foundation's `URL` [has been replaced](https://github.com/apple/swift-corelibs-foundation/pull/4955) with swift-foundation's.
I'd think that'll make it to Swift 6 but that's just my expectations.

It appears that swift-foundation has yet to implement `CharacterSet` and just falls back to the closed-source Darwin implementation for now, on Darwin.
Probably the best approach is the same as with `URL`: to implement `CharacterSet` in swift-foundation then replace corelibs-foundation's version with that. Though not sure if that's possible by Swift 6.

So we likely have to move on with some workarounds in OpenAPIRuntime for now, at least for `CharacterSet`.
I'd love to be proven wrong though.

leogdion:
> Can we embed the `nonisolated(unsafe)` section with `#if swift(>=6.0)` to allow to continue to use Swift < 6.0 toolchains ?

Done.

Sources/OpenAPIRuntime/Multipart/MultipartFramesToRawPartsSequence.swift
leogdion: Done. For the record, what's the difference between `compiler(>=6.0)` and `swift(>=6.0)`?

czechboy0:
I chatted with the Foundation folks and let's wait a week or two, their WIP PRs should land and have this fixed (Sendability of these two types).

Please separate the async sequence changes into a separate PR, we can land those.

leogdion:
> I chatted with the Foundation folks and let's wait a week or two, their WIP PRs should land and have this fixed (Sendability of these two types).
> 
> Please separate the async sequence changes into a separate PR, we can land those.

Done #108 

Any thoughts on how to move forward in the meantime regarding building my app in Swift 6?

czechboy0:
> Any thoughts on how to move forward in the meantime regarding building my app in Swift 6?

Yes, split out the generated code into a separate module/target, and use Swift 5 language mode on it. You can keep using Swift 6 in the rest of your codebase.

PR #109, opened 2024-07-17
Support NSNull in OpenAPIContainer types

czechboy0:
### Motivation

When receiving containers from adopter code, there might be NSNull values, which represent a nil value.

Previously, this value would not be treated as nil, instead it'd throw an error as an unrecognized type.

### Modifications

Handle NSNull and treat it as nil.

### Result

You can provide a container with an NSNull nested value and it'll get encoded correctly.

### Test Plan

Added a unit test.


simonjbeaumont (approved):
Nice catch.

PR #110, opened 2024-07-19
Improved encoding of NSNumber in OpenAPIValueContainer

czechboy0:
### Motivation

When getting CoreFoundation/Foundation types, especially numbers, they automatically bridge to Swift types like Bool, Int, etc. That casting is pretty flexible, and allows e.g. casting a number into a boolean, which isn't desired when encoding into JSON, as `false` and `0` represent very different values.

Previously, we relied on the automatic casting to know how to encode values, however that produced incorrect results in some cases.

### Modifications

Add explicit handling of CF/NS types and try to encode using that new method before falling back to testing for native Swift types. This ensures that the original intention of the creator of the CF/NS types doesn't get lost in encoding.

### Result

Correct encoding into JSON of types produced in the CF/NS world, like JSONSerialization.

### Test Plan

Added unit tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ DerivedData/
 /Package.resolved
 .ci/
 .docc-build/
+.swiftpm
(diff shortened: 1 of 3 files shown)


simonjbeaumont (approved):
Well... this is fun. Nice patch though.

PR #111, opened 2024-07-29
[Pitch] Custom JSON coder

czechboy0:
### Motivation

An idea of how to address https://github.com/apple/swift-openapi-generator/issues/605 in a more holistic way, following the approach we took with the date transcoder and XML custom coder.

### Modifications

- Reuses and extends (with backwards shims) the `CustomCoder` protocol.
- Adds a `jsonCoder` parameter to the `Configuration` struct.
- Ensures the date transcoder still gets applied on the JSON coder (even when a custom one is provided).
- Added convenience initializers for the case when the JSON encoder's output formatting is the only thing the user wants to customize (expected to be the vast majority of cases).

### Result

Fixes #605 while also providing a customization hook allowing a completely custom JSON encoder and decoder. This allows handling new JSON standard improvements that are not yet present in Foundation's JSON encoder/decoder, or even processing the JSON transparently (e.g. inline encryption).

### Test Plan

Added unit tests.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/Configuration.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Configuration.swift
@@ -96,29 +96,11 @@ extension JSONDecoder.DateDecodingStrategy {
     }
 }
 
-/// A type that allows custom content type encoding and decoding.
-public protocol CustomCoder: Sendable {
-
-    /// Encodes the given value and returns its custom encoded representation.
-    ///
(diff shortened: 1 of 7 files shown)


czechboy0:
Discussed more with @simonjbeaumont, this might be opening too much of an API surface for the ask. Will open a new PR with a solution that just allows providing the JSON encoding options, without having to provide a whole coder.

PR #112, opened 2024-07-29
Custom JSON encoding options

czechboy0:
### Motivation

A second attempt at https://github.com/apple/swift-openapi-generator/issues/605, this time less open-ended than #111.

Only allow customizing the JSON encoding options.

### Modifications

- Added a `JSONEncodingOptions` option set.
- Added a corresponding property on the `Configuration` struct.
- Piped the value to the underlying JSONEncoder.

### Result

Allows users to customize the JSON encoding options.

### Test Plan

Added unit tests.

### DIFF
--- a/Tests/OpenAPIRuntimeTests/Conversion/Test_Configuration.swift
+++ b/Tests/OpenAPIRuntimeTests/Conversion/Test_Configuration.swift
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 import XCTest
-@_spi(Generated) import OpenAPIRuntime
+@_spi(Generated) @testable import OpenAPIRuntime
 
 final class Test_Configuration: Test_Runtime {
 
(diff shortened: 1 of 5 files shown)


Tests/OpenAPIRuntimeTests/Conversion/Test_Configuration.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 import XCTest
-@_spi(Generated) import OpenAPIRuntime
+@_spi(Generated) @testable import OpenAPIRuntime
```
simonjbeaumont, line 15: OOI, why was the `@testable` import needed here? Could we write the tests to use only the API?

simonjbeaumont (approved):
Thanks @czechboy0, LGTM.

I think this is probably the most pragmatic approach for now. We can always revisit a full-blown encoder abstraction in the future when there's a real need for it.

Tests/OpenAPIRuntimeTests/Conversion/Test_Configuration.swift
czechboy0: Without it, I can't access the `encoder` directly to try to encode a few sample values. Otherwise I'd have to go through the SPI for generated code, and those require creating a bit more state to invoke.

It's intentional, we _want_ to hide `Foundation.JSONEncoder` from the API (and SPI), but it does require us to either use `@testable` here or make the test go through a bit more unrelated code.

simonjbeaumont: I should have been clearer. I could see what was using the `@testable` import, I was asking why we didn't exercise this new functionality through the public API only, which I've been leaning to these days because it prevents accidentally releasing things that pass tests but haven't been exposed properly in the API.

For clarity this is not a blocking comment and I'm happy to stick with the way it is.

czechboy0: No you bring up a good point, let me update and drop the `@testable` import.

czechboy0: Ok updated.

simonjbeaumont (approved):

PR #113, opened 2024-08-13
[Runtime] Improve parameter handling of MIME types in content types

czechboy0:
### Motivation

In service of an upcoming generator PR - to better handle parameters in MIME types.

This is needed for APIs such as Kubernetes, that has both `application/json` and `application/json; watch=true` in a single operation.

### Modifications

Use the more modern `OpenAPIMIMEType` type for parsing and comparing content types, rather than the old naive logic that ignored parameter (mis)matches.

### Result

More accurate handling of content types.

### Test Plan

Adapted unit tests.


Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
```diff
@@ -56,11 +56,34 @@ extension Converter {
...
+            return
+        }
+
+        // Look for the first match.
+        for acceptValue in acceptValues {
+            // Fast path.
+            if acceptValue == substring { return }
+            guard let parsedAcceptValue = OpenAPIMIMEType(acceptValue) else {
+                throw RuntimeError.invalidExpectedContentType(acceptValue)
+            }
```
simonjbeaumont, line 75: How permissive would we like to be here? If a client sends an `Accept` header with a valid content-type that the server supports, and an invalid one, should the server permit it? IIUC, this `guard` statement suggests not, but is that the desired behaviour?

simonjbeaumont (changes requested):

Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
czechboy0: If we fail to even parse the header value in its entirety, I suspect something very wrong is happening and I think we should fail, yeah. If we skip it silently, it could lead to harder-to-debug cases where the user is getting unexpected content back, and we're not catching that on the server (even though we could).

Or do you have cases in mind where this could be too strict?

Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
```diff
@@ -56,11 +56,34 @@ extension Converter {
...
-        if acceptValues.contains("\(substring.split(separator: "/")[0].lowercased())/*") { return }
-        if acceptValues.contains(where: { $0.localizedCaseInsensitiveContains(substring) }) { return }
+        guard let parsedSubstring = OpenAPIMIMEType(substring) else {
+            throw RuntimeError.invalidAcceptSubstring(substring)
+        }
+        guard case .concrete(let substringType, let substringSubtype) = parsedSubstring.kind else {
+            // If the substring content type has a wildcard, just let it through.
+            // It's not well defined how such a case should behave, so be permissive.
+            return
+        }
```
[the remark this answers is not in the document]
czechboy0: Yeah this case isn't handled correctly in this shape, I tried to explain that in the comment in the `else` block. In general we don't have clarity on what to do with e.g. `application/*` (a partial wildcard) being the _content type_ in the OpenAPI doc (as opposed to wildcards being part of the Accept header, which is well-defined). That's why I opted to go more relaxed here, as I don't know what to back up with being stricter.

I guess it comes down to whether we think that `application/*` as a content type that we have any information about whatsoever (so far I've though the answer is no, and treated it equally to `*/*`). Maybe that's not the right place to land, and we can come up with rules that constrain it more. We just run the risk of getting it slightly wrong at which point users won't have an escape hatch. But I could be convinced of going in either direction.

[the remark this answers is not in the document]
simonjbeaumont: I'd have thought that if we have a partial, we would validate that the top-level types match in the validate function.

I can see that a partial wildcard, e.g. `application/*`, is a strange thing to return in the `Content-Type` header field of a HTTP response, but it's clear that it doesn't match an `Accept` header of `text/plain`, or `text/*`, or anything that doesn't start with `application/`.

For this reason, I think the least surprising semantics of this validate function is to reject it.

To pull on this thread a little more though:

1. Is a partial wildcard content-type for a response valid OpenAPI (according to OAS)?
2. If yes, then is that just an oversight? I.e. if it is _valid_, is it typically always a _mistake_, or is there a genuine use case for it?

The answers to these two questions might yield some more validation and/or warning generation when the generator runs.

Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
simonjbeaumont: I think we need to consider the case where someone is calling the server from a not-generated client and might make mistakes and provide a malformed content-type followed by a valid one in the accept header. Having them get back a response of the valid one seems like a nicer experience than failing the request altogether, but I appreciate your opinion might differ here.

Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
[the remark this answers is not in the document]
czechboy0: OpenAPI does allow partial and full wildcards in the content type, but I've not come up with a way to treat that any differently to raw data that we know nothing about.

Agreed that we should do at least the type match. I'll update the PR.

Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
czechboy0: Yeah I genuinely wonder, if they're typing in a malformed content type, whether they'd want us to fall back to the other way, or to error out (since there's probably a reason they added it in the first place).

Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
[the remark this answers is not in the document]
simonjbeaumont: From chatting with folks in the team and consulting the RFCs, I think we need to adjust this differently.

For `Content-Type: a/*`, I suggested it should be permitted if `Accept: a/b` but we should only permit if `Accept: a/*` because we should treat `a/*` as a concrete MIME type with `*` being an unrecognised, but concrete (not wildcard) subtype, when in the `Content-Type` header field.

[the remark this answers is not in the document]
czechboy0: Ok, updated with this more correct matching. Check out the unit tests, I can add more if there's a case I missed.

Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
czechboy0: Discussed more offline, we'll leave it this strict, but separately I also filed https://github.com/apple/swift-openapi-generator/issues/644 so that we actually return 400, not 500 (as we'd return today when any error is thrown, including input validation error).

Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
```diff
@@ -56,11 +56,42 @@ extension Converter {
...
+            case (.concrete(type: let acceptType, subtype: let acceptSubtype), let substring):
+                if case let .concrete(substringType, substringSubtype) = substring {
+                    if acceptType.lowercased() == substringType.lowercased()
+                        && acceptSubtype.lowercased() == substringSubtype.lowercased()
+                    {
+                        return
+                    }
+                }
+            }
+        }
```
simonjbeaumont, line 94: Functionally I've convinced myself this is fine but I found the combination of doing all accept header values in the loop, the use of nested switch and the early-return to be a challenge to understand.

I'd really like to factor this out into a function that operates on two parsed `OpenAPIMIMETypes` and returns a boolean whether one can satisfy the other. 

Essentially, adding _this_ to `OpenAPIMIMEType`:

```swift
    func satisfies(parsedAcceptValue: OpenAPIMIMEType) -> Bool {
        switch (parsedAcceptValue.kind, self.kind) {
        case (.concrete, .any), (.concrete, .anySubtype), (.anySubtype, .any):
            // The response content-type must be at least as specific as the accept header.
            return false
        case (.any, _):
            // Accept: */* -- Any content-type satisfies the accept header.
            return true
        case (.anySubtype(let acceptType), .anySubtype(let substringType)),
            (.anySubtype(let acceptType), .concrete(let substringType, _)):
            // Accept: type/* -- The content-type should match the partially-specified accept header.
            return acceptType.lowercased() == substringType.lowercased()
        case (.concrete(let acceptType, let acceptSubtype), .concrete(let substringType, let substringSubtype)):
            // Accept: type/subtype -- The content-type should match the concrete type.
            return acceptType.lowercased() == substringType.lowercased() && acceptSubtype.lowercased() == substringSubtype.lowercased()
        }
    }
```

Then this function that does validation and throws can become:

```swift
...
        // Look for the first match.
        for acceptValue in acceptValues {
            // Fast path.
            if acceptValue == substring { return }
            guard let parsedAcceptValue = OpenAPIMIMEType(acceptValue) else {
                throw RuntimeError.invalidExpectedContentType(acceptValue)
            }
            if parsedSubstring.satisfies(parsedAcceptValue: parsedAcceptValue) { return }
        }
        throw RuntimeError.unexpectedAcceptHeader(acceptHeader)

```

It affords for a more testable, composable function too.

simonjbeaumont (changes requested):

Sources/OpenAPIRuntime/Conversion/Converter+Server.swift
czechboy0: Thanks, updated!

simonjbeaumont (approved):
Nice one, thanks!

PR #115, opened 2024-09-17
EventStreams: Customisable Terminating Byte Sequence

paulhdk:
### Motivation

As discussed in https://github.com/apple/swift-openapi-generator/issues/622, some APIs, e.g., ChatGPT or Claude, may return a non-JSON byte sequence to terminate a stream of events.
If not handled with a workaround (see below)such non-JSON terminating byte sequences cause a decoding error.

### Modifications

This PR adds the ability to customise the terminating byte sequence by providing a closure to `asDecodedServerSentEvents()` as well as `asDecodedServerSentEventsWithJSONData()` that can match incoming data for the terminating byte sequence before it is decoded into JSON, for instance.

### Result

Instead of having to decode and re-encode incoming events to filter out the terminating byte sequence - as seen in https://github.com/apple/swift-openapi-generator/issues/622#issuecomment-2346391088 - terminating byte sequences can now be cleanly caught by either providing a closure or providing the terminating byte sequence directly when calling `asDecodedServerSentEvents()` and `asDecodedServerSentEventsWithJSONData()`.

### Test Plan

This PR includes unit tests that test the new function parameters as part of the existing tests for `asDecodedServerSentEvents()` as well as `asDecodedServerSentEventsWithJSONData()`.

### DIFF
--- a/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
+++ b/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
@@ -28,9 +28,9 @@ where Upstream.Element == ArraySlice<UInt8> {
     /// The upstream sequence.
     private let upstream: Upstream
 
-    /// A closure that determines whether the given byte sequence is the terminating byte sequence defined by the API.
+    /// A closure that determines whether the given byte chunk should be forwarded to the consumer.
     /// - Parameter: A byte chunk.
-    /// - Returns: `True` until the terminating byte sequence is received.
+    /// - Returns: `true` if the byte chunk should be forwarded, `false` if this byte chunk is the terminating sequence.
     private let predicate: @Sendable (ArraySlice<UInt8>) -> Bool
 
     /// Creates a new sequence.
--- a/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
+++ b/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
@@ -58,7 +58,7 @@ extension ServerSentEventsDeserializationSequence: AsyncSequence {
         /// The state machine of the iterator.
         var stateMachine: StateMachine
 
-        /// A closure that determines whether the given byte sequence is the terminating byte sequence defined by the API.
+        /// A closure that determines whether the given byte chunk should be forwarded to the consumer.
         /// - Parameter: A byte chunk.
         /// - Returns: `True` until the terminating byte sequence is received.
         let predicate: (ArraySlice<UInt8>) -> Bool
--- a/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
+++ b/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
@@ -100,10 +100,11 @@ extension AsyncSequence where Element == ArraySlice<UInt8>, Self: Sendable {
     /// Use this method if the event's `data` field is not JSON, or if you don't want to parse it using `asDecodedServerSentEventsWithJSONData`.
     /// - Parameter: A closure that determines whether the given byte chunk should be forwarded to the consumer.
     /// - Returns: A sequence that provides the events.
-    public func asDecodedServerSentEvents(while predicate: @escaping @Sendable (ArraySlice<UInt8>) -> Bool = { _ in true }) -> ServerSentEventsDeserializationSequence<
-        ServerSentEventsLineDeserializationSequence<Self>
-    > { .init(upstream: ServerSentEventsLineDeserializationSequence(upstream: self), while: predicate) }
-    
+    public func asDecodedServerSentEvents(
+        while predicate: @escaping @Sendable (ArraySlice<UInt8>) -> Bool = { _ in true }
+    ) -> ServerSentEventsDeserializationSequence<ServerSentEventsLineDeserializationSequence<Self>> {
+        .init(upstream: ServerSentEventsLineDeserializationSequence(upstream: self), while: predicate)
+    }
     /// Returns another sequence that decodes each event's data as the provided type using the provided decoder.
     ///
     /// Use this method if the event's `data` field is JSON.
@@ -188,7 +189,6 @@ extension ServerSentEventsDeserializationSequence.Iterator {
                     // Dispatch the accumulated event.
                     // If the last character of data is a newline, strip it.
                     if event.data?.hasSuffix("\n") ?? false { event.data?.removeLast() }
-                    
                     if let data = event.data, !predicate(ArraySlice(data.utf8)) {
                         state = .finished
                         return .returnNil
--- a/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
+++ b/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
@@ -58,11 +58,6 @@ extension ServerSentEventsDeserializationSequence: AsyncSequence {
         /// The state machine of the iterator.
         var stateMachine: StateMachine
 
-        /// A closure that determines whether the given byte chunk should be forwarded to the consumer.
-        /// - Parameter: A byte chunk.
-        /// - Returns: `true` if the byte chunk should be forwarded, `false` if this byte chunk is the terminating sequence.
-        let predicate: (ArraySlice<UInt8>) -> Bool
-
         /// Creates a new sequence.
         /// - Parameters:
         ///     - upstream: The upstream sequence of arbitrary byte chunks.
@@ -70,7 +65,6 @@ extension ServerSentEventsDeserializationSequence: AsyncSequence {
         init(upstream: UpstreamIterator, while predicate: @escaping ((ArraySlice<UInt8>) -> Bool)) {
             self.upstream = upstream
             self.stateMachine = .init(while: predicate)
-            self.predicate = predicate 
         }
 
         /// Asynchronously advances to the next element and returns it, or ends the
--- a/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
+++ b/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
@@ -104,7 +104,7 @@ extension AsyncSequence where Element == ArraySlice<UInt8>, Self: Sendable {
     /// Returns another sequence that decodes each event's data as the provided type using the provided decoder.
     ///
     /// Use this method if the event's `data` field is not JSON, or if you don't want to parse it using `asDecodedServerSentEventsWithJSONData`.
-    /// - Parameter: A closure that determines whether the given byte sequence is the terminating byte sequence defined by the API.
+    /// - Parameter: A closure that determines whether the given byte chunk should be forwarded to the consumer.
     /// - Returns: A sequence that provides the events.
     public func asDecodedServerSentEvents(while predicate: @escaping @Sendable (ArraySlice<UInt8>) -> Bool = { _ in true }) -> ServerSentEventsDeserializationSequence<
         ServerSentEventsLineDeserializationSequence<Self>
--- a/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
+++ b/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
@@ -106,11 +106,6 @@ extension AsyncSequence where Element == ArraySlice<UInt8>, Self: Sendable {
         ServerSentEventsLineDeserializationSequence<Self>
     > { .init(upstream: ServerSentEventsLineDeserializationSequence(upstream: self), terminate: terminate) }
     
-    /// Convenience function for `asDecodedServerSentEvents` that directly receives the terminating byte sequence.
-    public func asDecodedServerSentEvents(terminatingSequence: ArraySlice<UInt8>) -> ServerSentEventsDeserializationSequence<
-        ServerSentEventsLineDeserializationSequence<Self>
-    > { asDecodedServerSentEvents(terminate: { incomingSequence in return incomingSequence == terminatingSequence }) }
-
     /// Returns another sequence that decodes each event's data as the provided type using the provided decoder.
     ///
     /// Use this method if the event's `data` field is JSON.
@@ -138,19 +133,6 @@ extension AsyncSequence where Element == ArraySlice<UInt8>, Self: Sendable {
                     retry: event.retry
                 )
             }
-    }
-
-    public func asDecodedServerSentEventsWithJSONData<JSONDataType: Decodable>(
-        of dataType: JSONDataType.Type = JSONDataType.self,
-        decoder: JSONDecoder = .init(),
-        terminatingData: ArraySlice<UInt8>
-    ) -> AsyncThrowingMapSequence<
-        ServerSentEventsDeserializationSequence<ServerSentEventsLineDeserializationSequence<Self>>,
-        ServerSentEventWithJSONData<JSONDataType>
-    > {
-        asDecodedServerSentEventsWithJSONData(of: dataType, decoder: decoder) { incomingData in
-            terminatingData == incomingData
-        }
     }
 }
 
--- a/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
+++ b/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
@@ -36,7 +36,7 @@ where Upstream.Element == ArraySlice<UInt8> {
     /// Creates a new sequence.
     /// - Parameters:
     ///     - upstream: The upstream sequence of arbitrary byte chunks.
-    ///     - while: A closure that determines whether the given byte chunk should be forwarded to the consumer.
+    ///     - predicate: A closure that determines whether the given byte chunk should be forwarded to the consumer.
     public init(upstream: Upstream, while predicate: @escaping @Sendable (ArraySlice<UInt8>) -> Bool) {
         self.upstream = upstream
         self.predicate = predicate
--- a/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
+++ b/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
@@ -166,7 +166,8 @@ extension ServerSentEventsDeserializationSequence.Iterator {
         /// Creates a new state machine.
         init(while predicate: @escaping (ArraySlice<UInt8>) -> Bool) {
             self.state = .accumulatingEvent(.init(), buffer: [])
-            self.predicate = predicate}
+            self.predicate = predicate
+        }
 
         /// An action returned by the `next` method.
         enum NextAction {
--- a/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
+++ b/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
@@ -60,7 +60,7 @@ extension ServerSentEventsDeserializationSequence: AsyncSequence {
 
         /// A closure that determines whether the given byte chunk should be forwarded to the consumer.
         /// - Parameter: A byte chunk.
-        /// - Returns: `True` until the terminating byte sequence is received.
+        /// - Returns: `true` if the byte chunk should be forwarded, `false` if this byte chunk is the terminating sequence.
         let predicate: (ArraySlice<UInt8>) -> Bool
 
         init(upstream: any AsyncIteratorProtocol, while predicate: @escaping ((ArraySlice<UInt8>) -> Bool)) {
--- a/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
+++ b/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
@@ -63,6 +63,10 @@ extension ServerSentEventsDeserializationSequence: AsyncSequence {
         /// - Returns: `true` if the byte chunk should be forwarded, `false` if this byte chunk is the terminating sequence.
         let predicate: (ArraySlice<UInt8>) -> Bool
 
+        /// Creates a new sequence.
+        /// - Parameters:
+        ///     - upstream: The upstream sequence of arbitrary byte chunks.
+        ///     - while: A closure that determines whether the given byte chunk should be forwarded to the consumer.
         init(upstream: UpstreamIterator, while predicate: @escaping ((ArraySlice<UInt8>) -> Bool)) {
             self.upstream = upstream
             self.stateMachine = .init(while: predicate)
--- a/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
+++ b/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
@@ -199,14 +199,14 @@ extension ServerSentEventsDeserializationSequence.Iterator {
                 buffer.removeFirst()
                 if line.isEmpty {
                     // Dispatch the accumulated event.
-                    state = .accumulatingEvent(.init(), buffer: buffer)
                     // If the last character of data is a newline, strip it.
                     if event.data?.hasSuffix("\n") ?? false { event.data?.removeLast() }
-
-                    if let data = event.data {
-                        if !predicate(ArraySlice(Data(data.utf8))) {
-                            return .returnNil
-                        }
+                    
+                    state = .accumulatingEvent(.init(), buffer: buffer)
+                    
+                    if let data = event.data, !predicate(ArraySlice(data.utf8)) {
+                        state = .finished
+                        return .returnNil
                     }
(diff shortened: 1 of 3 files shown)


simonjbeaumont (commented):
Just a courtesy note to acknowledge this PR—thanks for getting stuck in!

I’m out sick so might not get to this immediately. A quick glance: you’ll need to not remove the existing public initialiser (note that adding a parameter to a public method is technically API breaking because you can refer to functions by their signature.

czechboy0:
Thanks @paulhdk, the general shape of this is correct, let's just polish it up a bit together.

1. As Si mentioned, changing (even adding) function parameters is an API-breaking change, so we need to add the old variants to Deprecated.swift and mark it as deprecated (you'll see the many other deprecated methods in that name, just follow that pattern). That'll fix the first piece of feedback.
2. While the method you used to terminate the sequence should work, I think there's a simpler way to do this: `AsyncSequence` contains the [`prefix(while:)`](https://developer.apple.com/documentation/swift/asyncsequence/prefix(while:)-2xy95) method, which is basically exactly what we need - we want the sequence to be iterated _until_ a specific element is encountered, at which point we want to stop. That's equivalent to the `prefix(while:)` method, just with the test closure inverted (_stop_ when the element _matches_ a pattern). That means that you don't need to change the two implementation-only async sequences at all, you'll just need to take the input sequence and call the `prefix(while:)` method on it, and provide that. Let me try to add a suggestion on the diff to illustrate this better.
3. The naming of the parameter is interesting. `terminate:` is fine, but I think we can make it even clearer. `terminator` is used by [`print`](https://developer.apple.com/documentation/swift/print(_:separator:terminator:)/), but in that case, it _adds_ the terminator to the rendered string, whereas here the parameter dictates when to _stop_ iterating the elements. It might still work as a parameter name for the variant that provides `ArraySlice<UInt8>`, and `terminateOn:` for the `(ArraySlice<UInt8>) -> Bool` variant, but I'm happy for us to bikeshed that here a bit more, I hope we can make the name even better.

czechboy0 (changes requested):
See comment above

paulhdk:
> Just a courtesy note to acknowledge this PR—thanks for getting stuck in!

Thank you both for all your help along the way!

> I’m out sick so might not get to this immediately. A quick glance: you’ll need to not remove the existing public initialiser (note that adding a parameter to a public method is technically API breaking because you can refer to functions by their signature.

> Thanks @paulhdk, the general shape of this is correct, let's just polish it up a bit together.
> 1. As Si mentioned, changing (even adding) function parameters is an API-breaking change, so we need to add the old variants to Deprecated.swift and mark it as deprecated (you'll see the many other deprecated methods in that name, just follow that pattern). That'll fix the first piece of feedback.

Ohh, I wasn't aware. But hadn't we discussed as an initial requirement that this PR shouldn't change the existing API? I thought about providing the changes of this PR in separate functions but I wasn't sure how to avoid duplicate code in that case. Is there an elegant way to preserve the existing API with this PR's changes?

> 2. While the method you used to terminate the sequence should work, I think there's a simpler way to do this: `AsyncSequence` contains the [`prefix(while:)`](https://developer.apple.com/documentation/swift/asyncsequence/prefix(while:)-2xy95) method, which is basically exactly what we need - we want the sequence to be iterated _until_ a specific element is encountered, at which point we want to stop. That's equivalent to the `prefix(while:)` method, just with the test closure inverted (_stop_ when the element _matches_ a pattern). That means that you don't need to change the two implementation-only async sequences at all, you'll just need to take the input sequence and call the `prefix(while:)` method on it, and provide that. Let me try to add a suggestion on the diff to illustrate this better.

Not sure if I'm missing something, but I'm not seeing your comment in the code? I just see "See comment above" but is not attached to any line of code.

> 3. The naming of the parameter is interesting. `terminate:` is fine, but I think we can make it even clearer. `terminator` is used by [`print`](https://developer.apple.com/documentation/swift/print(_:separator:terminator:)/), but in that case, it _adds_ the terminator to the rendered string, whereas here the parameter dictates when to _stop_ iterating the elements. It might still work as a parameter name for the variant that provides `ArraySlice<UInt8>`, and `terminateOn:` for the `(ArraySlice<UInt8>) -> Bool` variant, but I'm happy for us to bikeshed that here a bit more, I hope we can make the name even better.

Hm. What about "processUntil: {}"? In that case we would have to invert the return value of the closure though.
Or maybe `terminatingByteReceived: {}`?

czechboy0:
> Ohh, I wasn't aware. But hadn't we discussed as an initial requirement that this PR shouldn't change the existing API? I thought about providing the changes of this PR in separate functions but I wasn't sure how to avoid duplicate code in that case. Is there an elegant way to preserve the existing API with this PR's changes?

The usual steps we take when replacing an API while preserving it in a deprecated form:
1. Move the existing function into Deprecated.swift
2. Mark it as deprecated
3. Copy it into the original location and make the modification (you've already done this)

We can usually manage to have the deprecated variant call _into_ the new variant. In this case, I believe we should be able to make the deprecated variant call the new variant with a `{ false }` closure for the terminator, meaning it'll never terminate prematurely and will always run through the full sequence.

> Not sure if I'm missing something, but I'm not seeing your comment in the code? I just see "See comment above" but is not attached to any line of code.

Apologies, I got distracted and forgot to add the comment 🤦‍♂️ Let me do that now.

> Hm. What about processUntil: {}? In that case we would have to invert the return value of the closure though.

Yeah I expect us to need to invert the value if we use `prefix(while:)` under the hood, but that's okay I think.

Since it's a closure with a boolean result, maybe it could be called `isLastElement: (ArraySlice<UInt8>) -> Bool = { false }`? Or `shouldStop`? I'll let @simonjbeaumont help us with the bikeshedding here 🙂

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
```diff
@@ -79,26 +100,34 @@ extension AsyncSequence where Element == ArraySlice<UInt8>, Self: Sendable {
     /// Returns another sequence that decodes each event's data as the provided type using the provided decoder.
     ///
     /// Use this method if the event's `data` field is not JSON, or if you don't want to parse it using `asDecodedServerSentEventsWithJSONData`.
+    /// - Parameter: An optional closure that determines whether the given byte sequence is the terminating byte sequence defined by the API.
     /// - Returns: A sequence that provides the events.
-    public func asDecodedServerSentEvents() -> ServerSentEventsDeserializationSequence<
+    public func asDecodedServerSentEvents(terminate: (@Sendable (ArraySlice<UInt8>) -> Bool)? = nil) -> ServerSentEventsDeserializationSequence<
         ServerSentEventsLineDeserializationSequence<Self>
-    > { .init(upstream: ServerSentEventsLineDeserializationSequence(upstream: self)) }
+    > { .init(upstream: ServerSentEventsLineDeserializationSequence(upstream: self), terminate: terminate) }
```
czechboy0, line 107:
```suggestion
    > { .init(upstream: ServerSentEventsLineDeserializationSequence(upstream: self).prefix(while: { !terminate($0) }) }
```

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
```diff
@@ -79,26 +100,34 @@ extension AsyncSequence where Element == ArraySlice<UInt8>, Self: Sendable {
     /// Returns another sequence that decodes each event's data as the provided type using the provided decoder.
     ///
     /// Use this method if the event's `data` field is not JSON, or if you don't want to parse it using `asDecodedServerSentEventsWithJSONData`.
+    /// - Parameter: An optional closure that determines whether the given byte sequence is the terminating byte sequence defined by the API.
     /// - Returns: A sequence that provides the events.
-    public func asDecodedServerSentEvents() -> ServerSentEventsDeserializationSequence<
+    public func asDecodedServerSentEvents(terminate: (@Sendable (ArraySlice<UInt8>) -> Bool)? = nil) -> ServerSentEventsDeserializationSequence<
```
czechboy0, line 105: A possible alternative, but I don't feel strongly either way, looking for feedback
```suggestion
    public func asDecodedServerSentEvents(terminate: @Sendable (ArraySlice<UInt8>) -> Bool = { _ in false }) -> ServerSentEventsDeserializationSequence<
```

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
```diff
@@ -110,6 +139,19 @@ extension AsyncSequence where Element == ArraySlice<UInt8>, Self: Sendable {
                 )
             }
     }
+
+    public func asDecodedServerSentEventsWithJSONData<JSONDataType: Decodable>(
+        of dataType: JSONDataType.Type = JSONDataType.self,
+        decoder: JSONDecoder = .init(),
+        terminatingData: ArraySlice<UInt8>
```
czechboy0, line 146: `lastElement` or `stopOn` are other alternatives

simonjbeaumont: I think we need to name this to indicate that it's cutting the sequence off. In practice it will be used to detect what will be the last element anyway. 

I'm actually wondering if we should try and fit more into the existing nomenclature here. e.g. could this be called `while:` (like `prefix(while:)`). I think that's pretty defensible and clear. The semantics are _while this is true, decode the events as JSON data_.

czechboy0: I see, and in this case, the bool would be flipped, so like this?

```swift
response.asDecodedServerSentEventsWithJSONData(of: ..., while: { $0 != ArraySlice("[DONE]".utf8) })
```

Yeah I like that, it allows us to just reinforce conventions from the standard library rather than invent our own.

paulhdk: `while:` is perfect!

But we haven't settled on a name for what is currently called `terminatingData:`. Maybe `untilReceived:`? Can you think of something that fits in well with existing standard library nomenclature here as well?

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
paulhdk: I like this a lot better! It gets rid of an `if let` later on.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
simonjbeaumont: Is this for the overload that just takes the String/Data? I'm not sure about how to make that clear. Do we need to provide it all?

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
paulhdk: I'm getting 

> Cannot convert value of type 'AsyncPrefixWhileSequence<ServerSentEventsLineDeserializationSequence<Self>>' to expected argument type 'ServerSentEventsLineDeserializationSequence<Self>'

with

```swift
> { .init(upstream: ServerSentEventsLineDeserializationSequence(upstream: self).prefix(while: { if let terminate { return terminate($0) } else { return false } })) }
```

Should I be making the `prefix` call on the `upstream` paramter instead? Although that doesn't appear to work either.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
paulhdk: Yes, it's for the overload. @czechboy0, I believe you had suggested providing an additional overload initially. What do you reckon?

czechboy0: Yeah I suggested the overload, but I also don't have a good idea for the name, so let's hold off on adding it for now, and let's just do the closure-based variant for now?

czechboy0: I'd lean towards a single function that has a non-optional closure with a default value of `{ true }`. Basing this on e.g. `print(..., terminator:)`, which is also a non-optional defaulted value.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
czechboy0: Oh of course, I forgot we're not returning `some AsyncSequence<...>`, hmm.

I wonder if we should only add this for Swift 6 compiler and higher with `#if compiler(>=6)` (which doesn't mean we need Swift 6 language mode, however).

Otherwise we'll really need to pass this parameter into `ServerSentEventsLineDeserializationSequence` and make it implement the stopping (which is also not the worst outcome).

Thoughts, @simonjbeaumont?

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
paulhdk: Pushed. Does that work for you, @simonjbeaumont?

simonjbeaumont: Yep, I think I'd like the non-optional parameter here. When it was an early-terminator, I think optional made sense. Now it's essentially a filter, I think non-optional default to `{ true }` is better, unless there's no performance downsides.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
simonjbeaumont: My hunch is to _not_ make this conditional on a newer compiler because we'll just push the burden to adopters. It's fine when the adopter is a leaf package, but if they want to provide a library that supports older compilers they'll have to do a conditional-compilation dance too.

czechboy0: Yeah agreed - let just sink down the closure where it needs to be.

paulhdk: > Yeah agreed - let just sink down the closure where it needs to be.

I.e. keep it the way it is in the PR now?

paulhdk:
@czechboy0 @simonjbeaumont just giving this a polite bump 🙂

I believe I’ve addressed all of your feedback so far. I just need clarification on one point above, @czechboy0.

Let me know if there’s anything missing.

czechboy0 (changes requested):
Thanks for the ping, @paulhdk - I took a more detailed look and suggested a few more changes, mainly around the details of the sequence's state machine.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
```diff
@@ -133,8 +157,16 @@ extension ServerSentEventsDeserializationSequence.Iterator {
         /// The current state of the state machine.
         private(set) var state: State
 
+        
+        /// A closure that determines whether the given byte sequence is the terminating byte sequence defined by the API.
+        /// - Parameter: A sequence of byte chunks.
+        /// - Returns: `True` until the terminating byte sequence is received.
+        let predicate: (ArraySlice<UInt8>) -> Bool
```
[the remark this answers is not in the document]
paulhdk: The closure will cause the `State` enum to lose its conformance to `Hashable`. I don’t think there is an easy workaround for this, is there?

Tests/OpenAPIRuntimeTests/EventStreams/Test_ServerSentEventsDecoding.swift
```diff
@@ -16,17 +16,19 @@ import XCTest
 import Foundation
 
 final class Test_ServerSentEventsDecoding: Test_Runtime {
-    func _test(input: String, output: [ServerSentEvent], file: StaticString = #filePath, line: UInt = #line)
+    func _test(input: String, output: [ServerSentEvent], file: StaticString = #filePath, line: UInt = #line, while predicate: (ArraySlice<UInt8>) -> Bool = { _ in true }, eventCountOffset: Int = 0)
```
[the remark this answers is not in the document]
paulhdk: I had initially tried it without `eventCountOffset `, but the numbers didn’t match for me. 

```swift
    try await _test(
            input: #"""
                data: hello
                data: world

                data: [DONE]

                data: hello2
                data: world2


                """#,
            output: [
                .init(data: "hello\nworld")
            ],
            while: { incomingData in
                incomingData != ArraySlice<UInt8>(Data("[DONE]".utf8))
            },
            eventCountOffset: -2
```

For instance, without `eventsOffset`, this test case will return 3 instead of 1.

paulhdk:
> Thanks for the ping, @paulhdk - I took a more detailed look and suggested a few more changes, mainly around the details of the sequence's state machine.

Thanks for the review, @czechboy0! I’ve addressed everything and ran into some questions. See above.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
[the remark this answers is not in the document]
czechboy0: Try to remove the conformance completely, it might only be used in tests. If it's used there, please add a manual implementation of Equatable/Hashable, that just ignores the closure in equality/hashing. I don't believe we ever use equality checking of the state in production code.

Tests/OpenAPIRuntimeTests/EventStreams/Test_ServerSentEventsDecoding.swift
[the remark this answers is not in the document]
czechboy0: I suspect that's because the predicate isn't actually passed into the tested function `asDecodedServerSentEvents` below

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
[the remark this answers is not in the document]
paulhdk: Brilliant! simply removing it appears to have worked.

Tests/OpenAPIRuntimeTests/EventStreams/Test_ServerSentEventsDecoding.swift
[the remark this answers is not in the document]
paulhdk: Oh, that one flew right past me. Working now!

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
```diff
@@ -46,7 +56,22 @@ extension ServerSentEventsDeserializationSequence: AsyncSequence {
         var upstream: UpstreamIterator
 
         /// The state machine of the iterator.
-        var stateMachine: StateMachine = .init()
+        var stateMachine: StateMachine
+
+        /// A closure that determines whether the given byte chunk should be forwarded to the consumer.
+        /// - Parameter: A byte chunk.
+        /// - Returns: `true` if the byte chunk should be forwarded, `false` if this byte chunk is the terminating sequence.
+        let predicate: (ArraySlice<UInt8>) -> Bool
```
czechboy0, line 64: Would it be possible to also remove the `predicate` property from the iterator type? I believe it should be enough for it to be owned by the state machine.

So it'd stay in the initializer of the iterator, but wouldn't be persisted as a property on the iterator, only forwarded to the state machine's iterator.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
```diff
@@ -79,26 +104,29 @@ extension AsyncSequence where Element == ArraySlice<UInt8>, Self: Sendable {
     /// Returns another sequence that decodes each event's data as the provided type using the provided decoder.
     ///
     /// Use this method if the event's `data` field is not JSON, or if you don't want to parse it using `asDecodedServerSentEventsWithJSONData`.
+    /// - Parameter: A closure that determines whether the given byte sequence is the terminating byte sequence defined by the API.
```
czechboy0, line 107:
```suggestion
    /// - Parameter: A closure that determines whether the given byte chunk should be forwarded to the consumer.
```

czechboy0 (commented):
Two more minor comments, otherwise lgtm

simonjbeaumont:
Just a courtesy note to all authors of open PRs. We have just cut over to GitHub Actions for CI so you will see different checks coming past. There is also a note in the `CONTRIBUTING.md` on how to run these locally: https://github.com/apple/swift-openapi-runtime/blob/main/CONTRIBUTING.md#run-ci-checks-locally.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
paulhdk: Oh, of course. Done!

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
```diff
@@ -28,9 +28,19 @@ where Upstream.Element == ArraySlice<UInt8> {
...
+    /// - Parameter: A byte chunk.
+    /// - Returns: `true` if the byte chunk should be forwarded, `false` if this byte chunk is the terminating sequence.
+    private let predicate: @Sendable (ArraySlice<UInt8>) -> Bool
+
     /// Creates a new sequence.
-    /// - Parameter upstream: The upstream sequence of arbitrary byte chunks.
-    public init(upstream: Upstream) { self.upstream = upstream }
+    /// - Parameters:
+    ///     - upstream: The upstream sequence of arbitrary byte chunks.
+    ///     - while: A closure that determines whether the given byte chunk should be forwarded to the consumer.
```
czechboy0, line 39:
```suggestion
    ///     - predicate: A closure that determines whether the given byte chunk should be forwarded to the consumer.
```

IIRC the parameter name (not the label) is what's used in docs - try to run the soundness script to verify.

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
```diff
@@ -156,20 +180,26 @@ extension ServerSentEventsDeserializationSequence.Iterator {
...
                     // If the last character of data is a newline, strip it.
                     if event.data?.hasSuffix("\n") ?? false { event.data?.removeLast() }
+                    
+                    state = .accumulatingEvent(.init(), buffer: buffer, predicate: predicate)
+                    
+                    if let data = event.data, !predicate(ArraySlice(data.utf8)) {
+                        state = .finished
+                        return .returnNil
+                    }
                     return .emitEvent(event)
```
czechboy0, line 198:
```suggestion
                    if let data = event.data, !predicate(ArraySlice(data.utf8)) {
                        state = .finished
                        return .returnNil
                    }
                    state = .accumulatingEvent(.init(), buffer: buffer, predicate: predicate)
                    return .emitEvent(event)
```

Minor - the state should only be assigned a single time in each pass.

czechboy0 (commented):
Okay, added a few more nits, but this is now ready for CI - kicking off. Great work!

Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift
czechboy0: Yeah confirmed by CI:

```
Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift:40:5: warning: [ValidateDocumentationComments] change the parameters of the documentation of 'init' to match its parameters
Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift:65:9: warning: [ValidateDocumentationComments] change the parameters of the documentation of 'init' to match its parameters
Sources/OpenAPIRuntime/EventStreams/ServerSentEventsDecoding.swift:116:5: warning: [ValidateDocumentationComments] change the parameters of the documentation of 'asDecodedServerSentEventsWithJSONData' to match its parameters
```

czechboy0:
You'll still need to rerun the CI - see the updated CONTRIBUTING.md on how to use the `act` tool to run the format action locally, and commit the changes

paulhdk:
> You'll still need to rerun the formatter - see the updated CONTRIBUTING.md on how to use the `act` tool to run the format action locally, and commit the changes

PTA.: Ran the formatter locally. All checks should pass now!

PSchmiedmayer:
Congratulations @paulhdk; exciting to see this move into SpeziLLM and the work you have been doing there! 🚀

Thank you @czechboy0 & @simonjbeaumont for the support, reviews, and input!

paulhdk:
> Thanks again for your contribution and for your patience as we've worked through all the details 👏 🙏

Thank you @czechboy0  and @simonjbeaumont for your help! It was fun working this one out and I learnt lots!

czechboy0:
> It was fun working this one out and I learnt lots!

Glad to hear it! If you're ever looking to contribute more, browse through https://github.com/apple/swift-openapi-generator/issues and comment on anything that catches your eye, and we can work together again 🙂

PR #116, opened 2024-10-02
ci: Migrate to GitHub Actions and reusable workflows, part one

simonjbeaumont:
### Motivation

Following on from the migration in Swift OpenAPI Generator, we need to migrate this repo to GitHub Actions based CI and the reusable workflows that are currently in the NIO repo. 

### Modifications

In order to bootstrap the migration, we need to merge the workflows in, otherwise we won't get any PR feedback on them while we get them ready. As a practical matter, they are all passing locally (verified) by `act` but it would be nice to stage these in so we can keep a green CI while we migrate and decommission the old CI.

So this disables the soundness checks for now, so we can then use a follow up PR to do the cut over with testing in the PR.

### Result

Old CI still working, new CI should start running in some capacity.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,43 @@
+name: PR
+
+on:
+    pull_request:
+      types: [opened, reopened, synchronize]
+
+jobs:
+    soundness:
+        name: Soundness
+        uses: apple/swift-nio/.github/workflows/soundness.yml@main
+        with:
+            api_breakage_check_enabled: true
+            broken_symlink_check_enabled: true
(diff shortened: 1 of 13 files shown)


czechboy0:
lgtm, could you update CONTRIBUTING.md before I approve this? Or is that coming in the cleanup PR?

czechboy0 (approved):
Thanks!

simonjbeaumont:
> lgtm, could you update CONTRIBUTING.md before I approve this? Or is that coming in the cleanup PR?

Thanks. This is coming with the final cutover since, until it's the one we use in CI, it's not the one folks should run locally :)

PR #119, opened 2024-10-02
benchmarks: Add subdirectory Benchmarks/ package

simonjbeaumont:
### Motivation

We'd like to be able to benchmark various parts of the runtime library to make targeted and measurable performance improvements.

### Modifications

- Add a subdirectory directory package with the package-benchmark plugin.
- Add a benchmark for `ISO8601DateTranscoder.encode(_:)`.
- Add temporary benchmarks of Foundation ISO8601 date encoding APIs to guide improvements to `ISO8601DateTranscoder`.

### Result

- Can now run benchmarks using `swift package --package-path Benchmarks/ benchmark`.

### Test Plan

There is no CI for this currently. That can follow, but we wanted the package plugin to unblock incoming updates to the date transcoder.


Benchmarks/Package.swift
```diff
@@ -0,0 +1,24 @@
...
+        .package(url: "https://github.com/ordo-one/package-benchmark.git", from: "1.22.0"),
+    ],
+    targets: [
+        .executableTarget(
+            name: "OpenAPIRuntimeBenchmarks",
+            dependencies: [
+                .product(name: "Benchmark", package: "package-benchmark"),
+                .product(name: "OpenAPIRuntime", package: "swift-openapi-runtime"),
+            ],
+            path: "Benchmarks/OpenAPIRuntimeBenchmarks",
```
czechboy0, line 18: Do we need this nested Benchmarks folder?

czechboy0: We can discuss separately.

czechboy0 (approved):

Benchmarks/Package.swift
simonjbeaumont: We don't need it. But let's see how the GHA rollout shakes out and decide if we need to align everything.

PR #120, opened 2024-10-03
Support nested arrays of primitive values inside of objects

czechboy0:
### Motivation

It's a useful pattern to define a single JSON schema for all your (e.g. query) parameters, and handle them as a single object in your code.

In OpenAPI, that'd be expressed like this, for example:

```yaml
# parameter
name: myParams
in: query
explode: true
style: form
schema:
  $ref: '#/components/schemas/QueryObject'

# schema
QueryObject:
  type: object
  properties:
    myString:
      type: string
    myList:
      type: array
      items:
      	type: string
```

Until now, the `myList` property would not be allowed, and would fail to serialize and parse, as arrays within objects were not allowed for `form` style parameters (used by query items, by default).

### Modifications

This PR extends the support of the `form` style to handle single nesting in the top level objects. It does _not_ add support for arbitrarily deep nesting.

As part of this work, we also now allow the `deepObject` style to do the same - use arrays nested in an object.

### Result

The useful pattern of having an array within a "params" object works correctly now.

### Test Plan

Added unit tests for all 4 components: encoder, decoder, serializer, and parser.

### DIFF
--- a/Sources/OpenAPIRuntime/URICoder/Common/URIEncodedNode.swift
+++ b/Sources/OpenAPIRuntime/URICoder/Common/URIEncodedNode.swift
@@ -47,6 +47,16 @@ enum URIEncodedNode: Equatable {
         /// A date value.
         case date(Date)
     }
+
+    /// A primitive value or an array of primitive values.
+    enum PrimitiveOrArrayOfPrimitives: Equatable {
+
+        /// A primitive value.
+        case primitive(Primitive)
+
+        /// An array of primitive values.
+        case arrayOfPrimitives([Primitive])
+    }
 }
 
 extension URIEncodedNode {
--- a/Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
+++ b/Sources/OpenAPIRuntime/URICoder/Parsing/URIParser.swift
@@ -241,7 +241,6 @@ extension URIParser {
                 appendPair(key, [value])
             }
         }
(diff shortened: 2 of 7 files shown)


Sources/OpenAPIRuntime/URICoder/Common/URIEncodedNode.swift
```diff
@@ -47,6 +47,16 @@ enum URIEncodedNode: Equatable {
...
     }
+
+    /// A primitive value or an array of primitive values.
+    enum PrimitiveOrArrayOfPrimitives: Equatable {
+
+        /// A primitive value.
+        case primitive(Primitive)
+
+        /// An array of primitive values.
+        case arrayOfPrimitives([Primitive])
```
simonjbeaumont, line 58: OOI, why do we need this additional nested enum. Can't we have this as a case in the outer enum?

simonjbeaumont (commented):
Thanks! One question about whether we need to break out this type.

Sources/OpenAPIRuntime/URICoder/Common/URIEncodedNode.swift
czechboy0: We could, but it'd force us to handle that case in several codepaths where we don't care about the distinction between "array" and "arrayOfPrimitives" (since most of this code does exhaustive matching of enum cases).

We also don't want to pass this enum value anywhere else, it's exclusively for the functions that only want to handle one of these two cases, verified at build time.

simonjbeaumont (approved):

PR #122, opened 2024-10-15
[CI] Update to Swift 6 CI

FranzBusch:
# Motivation

We just updated our CI matrix in NIO to only support 5.9, 5.10 and 6.

# Modification

This PR updates the trigger files in this repo. Since this repo was always 5.9+ this is easy.

# Result

Up to date CI

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -7,24 +7,17 @@ on:
 jobs:
     soundness:
         name: Soundness
-        uses: apple/swift-nio/.github/workflows/soundness.yml@main
+        uses: swiftlang/github-workflows/.github/workflows/soundness.yml@main
         with:
-            api_breakage_check_enabled: true
-            broken_symlink_check_enabled: true
-            docs_check_enabled: true
-            format_check_enabled: true
-            license_header_check_enabled: true
             license_header_check_project_name: "SwiftOpenAPIGenerator"
-            shell_check_enabled: true
-            unacceptable_language_check_enabled: true
 
     unit-tests:
         name: Unit tests
         uses: apple/swift-nio/.github/workflows/unit_tests.yml@main
         with:
-            linux_5_8_enabled: false
             linux_5_9_arguments_override: "--explicit-target-dependency-import-check error"
             linux_5_10_arguments_override: "--explicit-target-dependency-import-check error"
+            linux_6_0_arguments_override: "--explicit-target-dependency-import-check error"
             linux_nightly_6_0_arguments_override: "--explicit-target-dependency-import-check error"
             linux_nightly_main_enabled: false
 
@@ -34,7 +27,6 @@ jobs:
         with:
             name: "Integration test"
             matrix_linux_command: "apt-get update -yq && apt-get install -yq jq && ./scripts/run-integration-test.sh"
-            matrix_linux_5_8_enabled: false
             matrix_linux_nightly_main_enabled: false
 
     swift-6-language-mode:
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,35 +1,35 @@
 name: PR
 
 on:
-    pull_request:
-      types: [opened, reopened, synchronize]
+  pull_request:
+    types: [opened, reopened, synchronize]
 
 jobs:
-    soundness:
-        name: Soundness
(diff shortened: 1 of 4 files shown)


czechboy0:
@FranzBusch 

```
[Invalid workflow file: .github/workflows/pull_request.yml#L10](https://github.com/apple/swift-openapi-runtime/actions/runs/11345355766/workflow)
error parsing called workflow
".github/workflows/pull_request.yml"
-> "apple/swift-nio/.github/workflows/soundness.yml@main"
: failed to fetch workflow: workflow was not found.
```

.github/workflows/pull_request.yml
```diff
@@ -9,22 +9,15 @@ jobs:
...
         uses: apple/swift-nio/.github/workflows/soundness.yml@main
         with:
-            api_breakage_check_enabled: true
-            broken_symlink_check_enabled: true
-            docs_check_enabled: true
-            format_check_enabled: true
-            license_header_check_enabled: true
             license_header_check_project_name: "SwiftOpenAPIGenerator"
-            shell_check_enabled: true
-            unacceptable_language_check_enabled: true
```
simonjbeaumont, line 13: Why are we removing these?

.github/workflows/pull_request.yml
```diff
@@ -9,22 +9,15 @@ jobs:
         name: Soundness
         uses: apple/swift-nio/.github/workflows/soundness.yml@main
```
czechboy0, line 10: Should this be switched to the reusable one?

.github/workflows/pull_request.yml
czechboy0: I assume because they're on by default?

.github/workflows/scheduled.yml
```diff
@@ -9,9 +9,9 @@ jobs:
         name: Unit tests
         uses: apple/swift-nio/.github/workflows/unit_tests.yml@main
```
czechboy0, line 10: Same here? https://github.com/apple/swift-openapi-runtime/pull/122/files#r1801118837

.github/workflows/pull_request.yml
FranzBusch: Correct. The default is `true` so no reason to pass it additionally

FranzBusch:
@czechboy0 @simonjbeaumont the documentation pipeline seems to be failing since switching over to the reusable repo updated the toolchain for that workflow from 5.10 to 6.0. It looks like there are some new warnings now that need to be fixed. Could you help me with that?

.github/workflows/pull_request.yml
```diff
@@ -7,24 +7,17 @@ on:
 jobs:
```
czechboy0, line 7: Seems the indentation is off two lines higher, @FranzBusch

czechboy0:
> @czechboy0 @simonjbeaumont the documentation pipeline seems to be failing since switching over to the reusable repo updated the toolchain for that workflow from 5.10 to 6.0. It looks like there are some new warnings now that need to be fixed. Could you help me with that?

Yes let me open a separate PR making those fixes first.

czechboy0:
@FranzBusch https://github.com/apple/swift-openapi-runtime/pull/123

czechboy0:
@FranzBusch Actually, we have a chicken-and-egg problem, you need to bring this tiny diff into your PR instead, I can't merge mine with green CI anyway: https://github.com/apple/swift-openapi-runtime/pull/123

FranzBusch:
@czechboy0 Thanks for that. I slightly tweaked your commit to also avoid the `Return:` on the closure since that documents the wrong thing. I am wondering if that should actually be allowed.

czechboy0:
@FranzBusch So the YAML linter still fails: `PR / Soundness / YAML lint check (pull_request)`

FranzBusch:
@czechboy0 fixed!

czechboy0:
> @czechboy0 fixed!

I still see it as failing 😬

FranzBusch:
@czechboy0 Finally it's all green. Please do me the honours!

.github/workflows/pull_request.yml
simonjbeaumont: Heh, OK. We discussed why we wanted them in when we onboarded them in the earlier PR, but fine.

PR #123, opened 2024-10-16
Fix up doc comments

czechboy0:
### Motivation

Unblock https://github.com/apple/swift-openapi-runtime/pull/122#issuecomment-2415982172

### Modifications

Fixed up a doc comment.

### Result

No more warnings with Swift 6 compiler docc.

### Test Plan

Verified locally.


czechboy0:
Will be merged as part of #122.

PR #126, opened 2024-11-10
Add error handling middleware

gayathrisairam:
### Motivation

Implementation of [Improved error handling proposal](https://forums.swift.org/t/proposal-soar-0011-improved-error-handling/74736)

### Modifications

Added HTTPResponseConvertible protocol
Added ErrorHandlingMiddleware that converts errors confirming to HTTPResponseConvertible to a HTTP response.

### Result

The new middleware is an opt-in middleware. So there won't be any change to existing clients.
Clients who wish to have OpenAPI error handling can include the new error middleware in their application.

### Test Plan

Added E2E tests to test the new middleware.

### DIFF
--- a/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
+++ b/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
@@ -14,7 +14,7 @@
 
 import HTTPTypes
 
-/// An opt-in error Handling middleware that converts an error to a HTTP response.
+/// An opt-in error handling middleware that converts an error to a HTTP response.
 ///
 /// Inclusion of  ``ErrorHandlingMiddleware`` should be accompanied by confirming errors to ``HTTPResponseConvertible``  protocol.
 /// Undocumented errors, i.e, errors not confriming to ``HTTPResponseConvertible`` are converted to a response with 500 status code.
--- a/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
+++ b/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
@@ -73,6 +73,7 @@ public protocol HTTPResponseConvertible {
     /// HTTP status to return in the response.
     var httpStatus: HTTPResponse.Status { get }
     /// Headers to return in the response.
+    ///
     /// This is optional as default values are provided in the extension.
     var httpHeaderFields: HTTPTypes.HTTPFields { get }
     /// (Optional) The body of the response to return
--- a/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
+++ b/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
@@ -76,7 +76,7 @@ public protocol HTTPResponseConvertible {
     ///
     /// This is optional as default values are provided in the extension.
     var httpHeaderFields: HTTPTypes.HTTPFields { get }
-    /// (Optional) The body of the response to return
+    /// The body of the HTTP response.
     var httpBody: OpenAPIRuntime.HTTPBody? { get }
 }
 
--- a/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
+++ b/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
@@ -16,7 +16,7 @@ import HTTPTypes
 
 /// An opt-in error handling middleware that converts an error to a HTTP response.
 ///
-/// Inclusion of  ``ErrorHandlingMiddleware`` should be accompanied by confirming errors to ``HTTPResponseConvertible``  protocol.
+/// Inclusion of  ``ErrorHandlingMiddleware`` should be accompanied by conforming errors to the ``HTTPResponseConvertible``  protocol.
 /// Undocumented errors, i.e, errors not confriming to ``HTTPResponseConvertible`` are converted to a response with 500 status code.
 ///
 /// Example usage
--- a/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
+++ b/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
@@ -72,7 +72,7 @@ public protocol HTTPResponseConvertible {
 
     /// HTTP status to return in the response.
     var httpStatus: HTTPResponse.Status { get }
-    /// Headers to return in the response.
+    /// The HTTP headers of the response.
     ///
     /// This is optional as default values are provided in the extension.
     var httpHeaderFields: HTTPTypes.HTTPFields { get }
--- a/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
+++ b/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
@@ -14,7 +14,7 @@
 
 import HTTPTypes
 
-/// An opt-in error handling middleware that converts an error to a HTTP response.
+/// An opt-in error handling middleware that converts an error to an HTTP response.
 ///
 /// Inclusion of  ``ErrorHandlingMiddleware`` should be accompanied by conforming errors to the ``HTTPResponseConvertible``  protocol.
 /// Errors not conforming to ``HTTPResponseConvertible`` are converted to a response with the 500 status code.
--- a/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
+++ b/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
@@ -19,7 +19,7 @@ import HTTPTypes
 /// Inclusion of  ``ErrorHandlingMiddleware`` should be accompanied by conforming errors to the ``HTTPResponseConvertible``  protocol.
 /// Errors not conforming to ``HTTPResponseConvertible`` are converted to a response with the 500 status code.
 ///
-/// Example usage
+/// ## Example usage
 /// 1. Create an error type that conforms to HTTPResponseConvertible protocol:
 ///
 /// ```swift
--- a/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
+++ b/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftOpenAPIGenerator open source project
 //
-// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
+// Copyright (c) 2024 Apple Inc. and the SwiftOpenAPIGenerator project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -14,16 +14,14 @@
 
 import HTTPTypes
 
-/// Error Handling middleware that converts an error to a HTTP response
+/// An opt-in error Handling middleware that converts an error to a HTTP response.
 ///
-/// This is an opt-in middleware that adopters may choose to include to convert application errors
-/// to a HTTP Response
-///
-/// Inclusion of this ErrorHandling Middleware should be accompanied by confirming errors to
-/// HTTPResponseConvertible protocol. Only errors  confirming to HTTPResponseConvertible are converted to a HTTP response. Other errors are re-thrown from this middleware.
+/// Inclusion of  ``ErrorHandlingMiddleware`` should be accompanied by confirming errors to ``HTTPResponseConvertible``  protocol.
+/// Undocumented errors, i.e, errors not confriming to ``HTTPResponseConvertible`` are converted to a response with 500 status code.
 ///
 /// Example usage
-/// 1. Create an error type that conforms to HTTPResponseConvertible protocol
+/// 1. Create an error type that conforms to HTTPResponseConvertible protocol:
+///
 /// ```swift
 /// extension MyAppError: HTTPResponseConvertible {
 ///    var httpStatus: HTTPResponse.Status {
@@ -37,13 +35,15 @@ import HTTPTypes
 /// }
 /// ```
 ///
-/// 2. Opt in to the error middleware while registering the handler
+/// 2. Opt in to the ``ErrorHandlingMiddleware``  while registering the handler:
 ///
 /// ```swift
 /// let handler = try await RequestHandler()
 /// try handler.registerHandlers(on: transport, middlewares: [ErrorHandlingMiddleware()])
-///
-
+/// ```
+/// - Note: The placement of ``ErrorHandlingMiddleware`` in the middleware chain is important.
+/// It should be determined based on the specific needs of each application.
+/// Consider the order of execution and dependencies between middlewares.
 public struct ErrorHandlingMiddleware: ServerMiddleware {
     public func intercept(_ request: HTTPTypes.HTTPRequest,
                           body: OpenAPIRuntime.HTTPBody?,
@@ -57,29 +57,28 @@ public struct ErrorHandlingMiddleware: ServerMiddleware {
                 return (HTTPResponse(status: appError.httpStatus, headerFields: appError.httpHeaderFields),
                 appError.httpBody)
             } else {
-                throw error
+                return (HTTPResponse(status: .internalServerError), nil)
             }
         }
     }
 }
 
-/// Protocol used by ErrorHandling middleware to map an error to a HTTPResponse
-///
-/// Adopters who wish to convert their application error to a HTTPResponse
-/// should confirm their error(s) to this protocol
-
+/// Protocol used by ErrorHandling middleware to map an error to a HTTPResponse.
+/// Adopters who wish to convert their application error to a HTTPResponse should confirm their error(s) to this protocol.
 public protocol HTTPResponseConvertible {
-    /// HTTP status to return in the response
+
+    /// HTTP status to return in the response.
     var httpStatus: HTTPResponse.Status { get }
     
-    /// (Optional) Headers to return in the response
+    /// Headers to return in the response.
+    /// This is optional as default values are provided in the extension.
     var httpHeaderFields: HTTPTypes.HTTPFields { get }
     
     /// (Optional) The body of the response to return
     var httpBody: OpenAPIRuntime.HTTPBody? { get }
 }
 
-/// Extension to HTTPResponseConvertible to provide default values for certian fields
+/// Extension to HTTPResponseConvertible to provide default values for certian fields.
 public extension HTTPResponseConvertible {
     var httpHeaderFields: HTTPTypes.HTTPFields { [:] }
     var httpBody: OpenAPIRuntime.HTTPBody? { nil }
--- a/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
+++ b/Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
@@ -45,17 +45,20 @@ import HTTPTypes
 /// It should be determined based on the specific needs of each application.
 /// Consider the order of execution and dependencies between middlewares.
 public struct ErrorHandlingMiddleware: ServerMiddleware {
-    public func intercept(_ request: HTTPTypes.HTTPRequest,
-                          body: OpenAPIRuntime.HTTPBody?,
-                          metadata: OpenAPIRuntime.ServerRequestMetadata,
-                          operationID: String,
-                          next: @Sendable (HTTPTypes.HTTPRequest, OpenAPIRuntime.HTTPBody?, OpenAPIRuntime.ServerRequestMetadata) async throws -> (HTTPTypes.HTTPResponse, OpenAPIRuntime.HTTPBody?)) async throws -> (HTTPTypes.HTTPResponse, OpenAPIRuntime.HTTPBody?) {
-        do {
-            return try await next(request, body, metadata)
-        } catch let error as ServerError {
+    public func intercept(
+        _ request: HTTPTypes.HTTPRequest,
+        body: OpenAPIRuntime.HTTPBody?,
+        metadata: OpenAPIRuntime.ServerRequestMetadata,
+        operationID: String,
+        next: @Sendable (HTTPTypes.HTTPRequest, OpenAPIRuntime.HTTPBody?, OpenAPIRuntime.ServerRequestMetadata)
+            async throws -> (HTTPTypes.HTTPResponse, OpenAPIRuntime.HTTPBody?)
+    ) async throws -> (HTTPTypes.HTTPResponse, OpenAPIRuntime.HTTPBody?) {
+        do { return try await next(request, body, metadata) } catch let error as ServerError {
             if let appError = error.underlyingError as? (any HTTPResponseConvertible) {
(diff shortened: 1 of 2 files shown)


Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+                throw error
+            }
+        }
+    }
+}
+
+/// Protocol used by ErrorHandling middleware to map an error to a HTTPResponse
+///
+/// Adopters who wish to convert their application error to a HTTPResponse
+/// should confirm their error(s) to this protocol
```
gayathrisairam, line 69: Do you prefer if this goes to a separate file?

Tests/OpenAPIRuntimeTests/Interface/Test_ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,144 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import HTTPTypes
+
+import XCTest
+@_spi(Generated) @testable import OpenAPIRuntime
+
+
```
gayathrisairam, line 20: I have added tests for testing the new middleware in isolation. 
I thought of adding UniversalServerTests with a mock handler and this error middleware. But that becomes more of an integration test. Let me know if you prefer them to be added.

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+/// ```
+///
+/// 2. Opt in to the error middleware while registering the handler
+///
+/// ```swift
+/// let handler = try await RequestHandler()
+/// try handler.registerHandlers(on: transport, middlewares: [ErrorHandlingMiddleware()])
+///
+
+public struct ErrorHandlingMiddleware: ServerMiddleware {
```
simonjbeaumont, line 47: Thank you so much for adding documentation comments; we really value them! Do you mind polishing them up a little, though?

- Ending sentences with punctuation: either full stop or colon, depending on context.
- Reflowing lines at a consistent width.
- The second sentence seems a little redundant given the headline comment.
- In the third, can we remove the double-space in the middle of the sentence, and use double-backticks to link to the referred documentation of `HTTPResponseConvertible`?
- Terminate the final code-fence.
- Remove any empty comment lines at the end of the doc comment block.
- Remove any empty lines between the comment block and the symbol to which it should be attached.

This feedback applies to any other documentation comments too.

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+public protocol HTTPResponseConvertible {
+    /// HTTP status to return in the response
+    var httpStatus: HTTPResponse.Status { get }
+    
+    /// (Optional) Headers to return in the response
+    var httpHeaderFields: HTTPTypes.HTTPFields { get }
+    
+    /// (Optional) The body of the response to return
+    var httpBody: OpenAPIRuntime.HTTPBody? { get }
+}
```
simonjbeaumont, line 80: I'll go and double check the proposal myself but I thought we were going to make these all optional? Right now, we have all three flavours:
- `httpStatus` is non-optional, with no default implementation;
- `httpHeaderFields` is non-optional, but has a default implementation; and
- `httpBody` is optional, with a default implementation.

IMO, if we're providing a default implementation in a protocol extension, we should _not_ use optional for the requirement.

I'll go double-check the proposal again, but, regardless of API spelling, is this the semantics we wanted here?

Specifically, if someone adds the `ErrorHandlingMiddleware`, don't we want it transform errors that do not conform into `500` and, in debug mode, possibly include something in the header or body?

Tests/OpenAPIRuntimeTests/Interface/Test_ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,144 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftOpenAPIGenerator open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftOpenAPIGenerator project authors
```
simonjbeaumont, line 5:
```suggestion
// Copyright (c) 2024 Apple Inc. and the SwiftOpenAPIGenerator project authors
```

Tests/OpenAPIRuntimeTests/Interface/Test_ErrorHandlingMiddleware.swift
simonjbeaumont: Adding these tests are great. IMO we don't need an integration test for this.

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
simonjbeaumont: Separately, should we add documentation about the importance of its position in the responder chain?

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
simonjbeaumont, line 1: General note: I've approved the CI run now, but you should probably try and run the soundness checks locally. See the contributing docs for that. This will allow you to at least run the formatter, which I imagine will be failing right now.

simonjbeaumont (commented):
Thanks for picking this up @gayathrisairam!

I've left some initial comments / questions, and a note on how you can run the linters locally.

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
czechboy0: > IMO, if we're providing a default implementation in a protocol extension, we should not use optional for the requirement.

I'm not sure I agree, I think it's good the way it is - header fields is a container that can be empty, so it's non-optional and defaults to empty in the protocol extension. The body, however, is a stateful async sequence, and the middleware API allows returning `nil`, so I think it should be optional and default to `nil` in the protocol extension.

> Specifically, if someone adds the ErrorHandlingMiddleware, don't we want it transform errors that do not conform into 500 and, in debug mode, possibly include something in the header or body?

Yes I also caught this one, I believe [here](https://forums.swift.org/t/proposal-soar-0011-improved-error-handling/74736/28) we agreed to always return a response, and use 500 for unconformed errors.

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
czechboy0: I'm fine with it staying all in the same file.

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
simonjbeaumont: > > IMO, if we're providing a default implementation in a protocol extension, we should not use optional for the requirement.
> 
> I'm not sure I agree, I think it's good the way it is - header fields is a container that can be empty, so it's non-optional and defaults to empty in the protocol extension. The body, however, is a stateful async sequence, and the middleware API allows returning `nil`, so I think it should be optional and default to `nil` in the protocol extension.

Right I think the way it's currently structured and documented is conflating two things. The optionality of whether a user needs to provide an override, and the optionality of the underlying field. We should have these match the target response type in terms of optionality, which, as you point out, they do already. Concretely, the `HTTPResponse` has non-optional status and headers (although headers may be empty), and an _optional_ body.

I think what's confusing then is the use of "(Optional)" in the documentation comments for the protocol requirements. How about we remove that term from the documentation and provide defaults for everything in the protocol extension to make it clear, including `httpBody`, which will default to `nil` and make it clear what the default is in the protocol level documentation?

gayathrisairam: You are right. I will convert undocumented errors to a 500 response. 
What do we want to do with header and body fields? Should we preserve the additional information in `ServerError.underlyingError` in the body as an application/json?

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
gayathrisairam: I have run the formatter and updated the PR with changes.

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+/// Inclusion of  ``ErrorHandlingMiddleware`` should be accompanied by confirming errors to ``HTTPResponseConvertible``  protocol.
+/// Undocumented errors, i.e, errors not confriming to ``HTTPResponseConvertible`` are converted to a response with 500 status code.
+///
+/// Example usage
+/// 1. Create an error type that conforms to HTTPResponseConvertible protocol:
+///
+/// ```swift
+/// extension MyAppError: HTTPResponseConvertible {
+///    var httpStatus: HTTPResponse.Status {
+///    switch self {
```
czechboy0, line 28:
```suggestion
///        switch self {
```

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+///            .forbidden
+///        }
+///    }
+/// }
+/// ```
+///
+/// 2. Opt in to the ``ErrorHandlingMiddleware``  while registering the handler:
+///
+/// ```swift
+/// let handler = try await RequestHandler()
```
czechboy0, line 41:
```suggestion
/// let handler = RequestHandler()
```

We can just show a simpler example, to avoid people assuming their handlers are expected/required to use `try await` for initializers.

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+/// try handler.registerHandlers(on: transport, middlewares: [ErrorHandlingMiddleware()])
+/// ```
+/// - Note: The placement of ``ErrorHandlingMiddleware`` in the middleware chain is important.
+/// It should be determined based on the specific needs of each application.
+/// Consider the order of execution and dependencies between middlewares.
+public struct ErrorHandlingMiddleware: ServerMiddleware {
+    public func intercept(
```
czechboy0, line 48:
```suggestion
public struct ErrorHandlingMiddleware: ServerMiddleware {

    public init() {}

    public func intercept(
```

We need a public initializer here.

And please add doc comments to both.

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+                    appError.httpBody
+                )
+            } else {
+                return (HTTPResponse(status: .internalServerError), nil)
+            }
+        }
+    }
+}
+
+/// Protocol used by ErrorHandling middleware to map an error to a HTTPResponse.
```
czechboy0, line 69:
```suggestion
/// A protocol used by ``ErrorHandlingMiddleware`` to map an error to an `HTTPResponse` and ``HTTPBody``.
```

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+/// Adopters who wish to convert their application error to a HTTPResponse should confirm their error(s) to this protocol.
```
czechboy0, line 70:
```suggestion
/// Adopters who wish to convert their application error to an `HTTPResponse` and ``HTTPBody`` should conform the error type to this protocol.
```

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+public protocol HTTPResponseConvertible {
+
+    /// HTTP status to return in the response.
```
czechboy0, line 73:
```suggestion
    /// An HTTP status to return in the response.
```

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+    var httpStatus: HTTPResponse.Status { get }
+    /// Headers to return in the response.
+    /// This is optional as default values are provided in the extension.
```
czechboy0, line 76:
```suggestion
    ///
    /// This is optional as default values are provided in the extension.
```

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+    var httpHeaderFields: HTTPTypes.HTTPFields { get }
+    /// (Optional) The body of the response to return
```
czechboy0, line 78:
```suggestion
    /// The body of the HTTP response.
```

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+/// Protocol used by ErrorHandling middleware to map an error to a HTTPResponse.
+/// Adopters who wish to convert their application error to a HTTPResponse should confirm their error(s) to this protocol.
+public protocol HTTPResponseConvertible {
+
+    /// HTTP status to return in the response.
+    var httpStatus: HTTPResponse.Status { get }
+    /// Headers to return in the response.
```
czechboy0, line 75:
```suggestion
    /// The HTTP headers of the response.
```

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,86 @@
...
+    var httpHeaderFields: HTTPTypes.HTTPFields { get }
+    /// (Optional) The body of the response to return
+    var httpBody: OpenAPIRuntime.HTTPBody? { get }
+}
+
+/// Extension to HTTPResponseConvertible to provide default values for certian fields.
+public extension HTTPResponseConvertible {
+    var httpHeaderFields: HTTPTypes.HTTPFields { [:] }
+    var httpBody: OpenAPIRuntime.HTTPBody? { nil }
+}
```
czechboy0, line 86:
```suggestion
/// Extension to HTTPResponseConvertible to provide default values for certain fields.
extension HTTPResponseConvertible {
    public var httpHeaderFields: HTTPTypes.HTTPFields { [:] }
    public var httpBody: OpenAPIRuntime.HTTPBody? { nil }
}
```

Tests/OpenAPIRuntimeTests/Interface/Test_ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,141 @@
...
+    var httpBody: OpenAPIRuntime.HTTPBody? = TEST_HTTP_BODY
+}
+
+struct PartialConvertibleError: Error, HTTPResponseConvertible {
+    var httpStatus: HTTPTypes.HTTPResponse.Status = HTTPResponse.Status.badRequest
+}
+
+struct NonConvertibleError: Error {}
+
+let TEST_HTTP_BODY = HTTPBody(try! JSONEncoder().encode(["error", " test error"]))
```
czechboy0, line 141:
```suggestion
let testHTTPBody = HTTPBody(try! JSONEncoder().encode(["error", " test error"]))
```

czechboy0 (changes requested):
Looks great overall, added a few suggestions for more polish, mainly in the doc comments.

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
simonjbeaumont: There's precedent for this being done in debug and not in release. And I'm personally in favour of having this behaviour. Not having it is quite an annoying developer experience IMO.

I'd also propose making it configurable and defaulting to off. But it should be a backwards compatible change so happy to defer it altogether for now.

czechboy0: I'd prefer to leave it out of this PR, however I would be supportive of a subsequent PR adding it as an explicit option. I really dislike the behavior being controlled by debug vs release.

simonjbeaumont: OK, @gayathrisairam feel free to break this out into a new GitHub issue where we can flesh out how this would look.

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,98 @@
...
+    }
+}
+
+/// A protocol used by ``ErrorHandlingMiddleware`` to map an error to an ``HTTPResponse`` and ``HTTPBody``.
+/// Adopters who wish to convert their application error to an `HTTPResponse` and ``HTTPBody`` should conform the error type to this protocol.
+public protocol HTTPResponseConvertible {
+
+    /// An HTTP status to return in the response.
+    var httpStatus: HTTPResponse.Status { get }
+    /// The HTTP headers of the response.
```
czechboy0, line 87:
```suggestion
    var httpStatus: HTTPResponse.Status { get }

    /// The HTTP header fields of the response.
```

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,98 @@
...
+    /// This is optional as default values are provided in the extension.
+    var httpHeaderFields: HTTPTypes.HTTPFields { get }
+    /// The body of the HTTP response.
```
czechboy0, line 90:
```suggestion
    var httpHeaderFields: HTTPTypes.HTTPFields { get }

    /// The body of the HTTP response.
```

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,98 @@
...
+    var httpBody: OpenAPIRuntime.HTTPBody? { get }
+}
+
+/// Extension to HTTPResponseConvertible to provide default values for certain fields.
+extension HTTPResponseConvertible {
+    public var httpHeaderFields: HTTPTypes.HTTPFields { [:] }
+    public var httpBody: OpenAPIRuntime.HTTPBody? { nil }
+}
```
czechboy0, line 98:
```suggestion
extension HTTPResponseConvertible {

    // swift-format-ignore: AllPublicDeclarationsHaveDocumentation
    public var httpHeaderFields: HTTPTypes.HTTPFields { [:] }

    // swift-format-ignore: AllPublicDeclarationsHaveDocumentation
    public var httpBody: OpenAPIRuntime.HTTPBody? { nil }
}
```

Tests/OpenAPIRuntimeTests/Interface/Test_ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,141 @@
...
+        let response = try await Test_ErrorHandlingMiddlewareTests.errorHandlingMiddleware.intercept(
+            Test_ErrorHandlingMiddlewareTests.mockRequest,
+            body: Test_ErrorHandlingMiddlewareTests.mockBody,
+            metadata: .init(),
+            operationID: "testop",
+            next: getNextMiddleware(failurePhase: .never)
+        )
+        XCTAssertEqual(response.0.status, .ok)
+    }
+    func testError_confirmingToProtocol_convertedToResponse() async throws {
```
czechboy0, line 35:
```suggestion
    }

    func testError_conformingToProtocol_convertedToResponse() async throws {
```

Tests/OpenAPIRuntimeTests/Interface/Test_ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,141 @@
...
+            metadata: .init(),
+            operationID: "testop",
+            next: getNextMiddleware(failurePhase: .convertibleError)
+        )
+        XCTAssertEqual(response.status, .badGateway)
+        XCTAssertEqual(response.headerFields, [.contentType: "application/json"])
+        XCTAssertEqual(responseBody, TEST_HTTP_BODY)
+    }
+
+    func testError_confirmingToProtocolWithoutAllValues_convertedToResponse() async throws {
```
czechboy0, line 48:
```suggestion
    func testError_conformingToProtocolWithoutAllValues_convertedToResponse() async throws {
```

Tests/OpenAPIRuntimeTests/Interface/Test_ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,141 @@
...
+            body: Test_ErrorHandlingMiddlewareTests.mockBody,
+            metadata: .init(),
+            operationID: "testop",
+            next: getNextMiddleware(failurePhase: .partialConvertibleError)
+        )
+        XCTAssertEqual(response.status, .badRequest)
+        XCTAssertEqual(response.headerFields, [:])
+        XCTAssertEqual(responseBody, nil)
+    }
+    func testError_notConfirmingToProtocol_returns500() async throws {
```
czechboy0, line 60:
```suggestion
    }

    func testError_notConformingToProtocol_returns500() async throws {
```

Tests/OpenAPIRuntimeTests/Interface/Test_ErrorHandlingMiddleware.swift
```diff
@@ -0,0 +1,141 @@
...
+            body: Test_ErrorHandlingMiddlewareTests.mockBody,
+            metadata: .init(),
+            operationID: "testop",
+            next: getNextMiddleware(failurePhase: .nonConvertibleError)
+        )
+        XCTAssertEqual(response.status, .internalServerError)
+        XCTAssertEqual(response.headerFields, [:])
+        XCTAssertEqual(responseBody, nil)
+    }
+    private func getNextMiddleware(failurePhase: MockErrorMiddleware_Next.FailurePhase) -> @Sendable (
```
czechboy0, line 72:
```suggestion
    }

    private func getNextMiddleware(failurePhase: MockErrorMiddleware_Next.FailurePhase) -> @Sendable (
```

czechboy0 (changes requested):
Thanks for the fixes, added a few more minor suggestions as part of the final pass

czechboy0:
@gayathrisairam please rerun the formatter and unit tests, those are failing

Sources/OpenAPIRuntime/Interface/ErrorHandlingMiddleware.swift
czechboy0: https://github.com/apple/swift-openapi-generator/issues/677

czechboy0:
Looks great, thank you @gayathrisairam! 👏

PR #127, opened 2024-11-13
Refactor URIDecoder/URIParser to improve handling of the deepObject style

czechboy0:
### Motivation

As part of https://github.com/apple/swift-openapi-generator/issues/259, adding deepObject parameter style support, the initial PR wasn't complete.

And once we dug more into it, turns out the original implementation of the URIDecoder/URIParser didn't really lend themselves well for handling deepObject, and the recent additions of supporting arrays within dictionaries (https://github.com/apple/swift-openapi-runtime/pull/120) further confused the implementation.

### Modifications

Refactored URIParser/URIDecoder with a clearer understanding of the current requirements. It's now much easier to follow and embraces the fact that each of the 7 variants of URI coding we support (form exploded, form unexploded, simple exploded, simple unexploded, form data exploded, form data unexploded, and now deepObject exploded) are similar, but still different in subtle ways.

This new implementation doesn't try as hard to share code between the implementations, so might at first sight appear to duplicate code. The original implementation had many methods with many configuration parameters and utility methods with a high cyclomatic complexity, which made it very hard to reason about. We did away with that.

While there, I also made some minor improvements to the serialization path, which allows cleaner round-tripping tests.

### Result

A more maintainable and more correct URI decoder/parser implementation.

### Test Plan

Added many more unit tests that test the full matrix of supported styles and inputs.


czechboy0:
PR is in draft as I need to add missing doc comments and a few polishing steps.

Sources/OpenAPIRuntime/URICoder/Decoding/URIValueFromNodeDecoder.swift
```diff
@@ -167,153 +135,240 @@ extension URIValueFromNodeDecoder {
...
+                result = .success(value)
+            } catch {
+                result = .failure(error)
+                throw error
             }
-            return []
+            cache.primitive = result
         }
-        return value
+        return try work(value)
```
simonjbeaumont, line 159: I'd like to see this pattern factored out as it's pretty repetitive.

I'm also wondering if `Optional<Result<>>` is the best representation for the state of the cache. I understand it fits with the three cases: (1) haven't tried, (2) tried and got a thing, and (3) tried and got an error; but I think an actual `enum CacheResult` with the three states in one enum might help us build some better layered logic.

simonjbeaumont (changes requested):
I'm really glad we're pushing this feature along. I've left some comments. Can I assume that all the new tests you added previously didn't work? If not, could you indicate which of them previously failed as I'd like to give those a little extra attention.

Sources/OpenAPIRuntime/URICoder/Decoding/URIValueFromNodeDecoder+Single.swift
```diff
@@ -33,7 +33,17 @@ extension URISingleValueDecodingContainer {
...
     private func _decodeBinaryFloatingPoint<T: BinaryFloatingPoint>(_: T.Type = T.self) throws -> T {
-        guard let double = try Double(value) else {
+        guard let value = try value else {
+            throw DecodingError.valueNotFound(
+                T.self,
+                DecodingError.Context.init(
+                    codingPath: codingPath,
+                    debugDescription: "Value not found.",
+                    underlyingError: nil
+                )
```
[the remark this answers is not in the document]
czechboy0: Agreed, I use .init when autocompletion isn't filtering enough and sometimes I forget to remove it later, thanks

Sources/OpenAPIRuntime/URICoder/Decoding/URIValueFromNodeDecoder+Unkeyed.swift
```diff
@@ -46,7 +36,7 @@ extension URIUnkeyedDecodingContainer {
     /// - Throws: An error if the container ran out of items.
     private mutating func _decodingNext<R>(in work: () throws -> R) throws -> R {
         guard !isAtEnd else { throw URIValueFromNodeDecoder.GeneralError.reachedEndOfUnkeyedContainer }
-        defer { values.formIndex(after: &index) }
+        defer { index += 1 }
```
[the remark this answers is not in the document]
czechboy0: It was, but we don't have access to the array in this container anymore, it's only present in the lower level decoder now as it gets parsed lazily.

Sources/OpenAPIRuntime/URICoder/Decoding/URIValueFromNodeDecoder+Unkeyed.swift
```diff
@@ -111,9 +101,9 @@ extension URIUnkeyedDecodingContainer {
 
 extension URIUnkeyedDecodingContainer: UnkeyedDecodingContainer {
 
-    var count: Int? { values.count }
+    var count: Int? { try? decoder.countOfCurrentArray() }
 
-    var isAtEnd: Bool { index == values.endIndex }
+    var isAtEnd: Bool { index == count }
 
     var currentIndex: Int { index }
```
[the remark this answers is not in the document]
czechboy0: Remnant of the older implementation, let me remove it

Sources/OpenAPIRuntime/URICoder/Decoding/URIValueFromNodeDecoder.swift
czechboy0: Factored the caching logic out and further simplified.

I kept the optional result type, as it gives us the nice `value.get()` "replay" ability of the cached value.

czechboy0:
All the previous tests passed previously, we just changed the tested abstractions which forced me to refractor the tests too. The only new functionality is the deepObject support, so tests related to that style are new.

Sources/OpenAPIRuntime/URICoder/Decoding/URIValueFromNodeDecoder.swift
```diff
@@ -345,16 +344,19 @@ extension URIValueFromNodeDecoder {
     /// contains a value for the provided key.
     /// - Parameter key: The key for which to look for a value.
     /// - Returns: `true` if a value was found, `false` otherwise.
-    /// - Throws: When parsing the root fails.
-    func containsElementInCurrentDictionary(forKey key: String) throws -> Bool {
-        try withCurrentDictionaryElements { dictionary in dictionary[key[...]] != nil }
+    func containsElementInCurrentDictionary(forKey key: String) -> Bool {
+        do { return try withCurrentDictionaryElements { dictionary in dictionary[key[...]] != nil } } catch {
+            return false
+        }
```
simonjbeaumont, line 350: nit: I still think if we're just swallowing the error and returning something else, then this is the nicer pattern, but it's not a blocker:
```suggestion
        (try? withCurrentDictionaryElements { dictionary in dictionary[key[...]] != nil }) ?? false
```

Sources/OpenAPIRuntime/URICoder/Decoding/URIValueFromNodeDecoder.swift
```diff
@@ -345,16 +344,19 @@ extension URIValueFromNodeDecoder {
...
 
     /// Returns a list of keys found in the current top-of-stack dictionary.
     /// - Returns: A list of keys from the dictionary.
     /// - Throws: When parsing the root fails.
-    func elementKeysInCurrentDictionary() throws -> [String] {
-        try withCurrentDictionaryElements { dictionary in dictionary.keys.map(String.init) }
+    func elementKeysInCurrentDictionary() -> [String] {
+        do { return try withCurrentDictionaryElements { dictionary in dictionary.keys.map(String.init) } } catch {
+            return []
+        }
```
simonjbeaumont, line 359: nit: I still think if we're just swallowing the error and returning something else, then this is the nicer pattern, but it's not a blocker:

```suggestion
        (try? withCurrentDictionaryElements { dictionary in dictionary.keys.map(String.init) } ) ?? []

```

simonjbeaumont (approved):
LGTM, thanks! I left one more comment but it's not a requirement.

Sources/OpenAPIRuntime/URICoder/Decoding/URIValueFromNodeDecoder.swift
czechboy0: Updated

Sources/OpenAPIRuntime/URICoder/Decoding/URIValueFromNodeDecoder.swift
czechboy0: Updated

PR #133, opened 2024-12-09
Enable MemberImportVisibility check on all targets

rnro:
Enable MemberImportVisibility check on all targets. Use a standard string header and footer to bracket the new block for ease of updating in the future with scripts.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -50,3 +50,14 @@ let package = Package(
         ),
     ]
 )
+
+// ---    STANDARD CROSS-REPO SETTINGS DO NOT EDIT   --- //
+for target in package.targets {
+    if target.type != .plugin {
+        var settings = target.swiftSettings ?? []
+        // https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md
+        settings.append(.enableUpcomingFeature("MemberImportVisibility"))
+        target.swiftSettings = settings
+    }
+}
+// --- END: STANDARD CROSS-REPO SETTINGS DO NOT EDIT --- //


simonjbeaumont:
Is this expected to be passing? 🤔

simonjbeaumont (approved):

Package.swift
```diff
@@ -23,30 +23,24 @@ let swiftSettings: [SwiftSetting] = [
...
-            name: "OpenAPIRuntime",
-            targets: ["OpenAPIRuntime"]
-        )
-    ],
-    dependencies: [
-        .package(url: "https://github.com/apple/swift-http-types", from: "1.0.0"),
-    ],
+    platforms: [.macOS(.v10_15), .macCatalyst(.v13), .iOS(.v13), .tvOS(.v13), .watchOS(.v6), .visionOS(.v1)],
+    products: [.library(name: "OpenAPIRuntime", targets: ["OpenAPIRuntime"])],
+    dependencies: [.package(url: "https://github.com/apple/swift-http-types", from: "1.0.0")],
```
czechboy0, line 28: Curious why this got reformatted now, even though the CI was green previously as well?

Package.swift
```diff
@@ -23,30 +23,24 @@ let swiftSettings: [SwiftSetting] = [
...
-            name: "OpenAPIRuntimeTests",
-            dependencies: ["OpenAPIRuntime"],
+            dependencies: [.product(name: "HTTPTypes", package: "swift-http-types")],
             swiftSettings: swiftSettings
-        ),
+        ), .testTarget(name: "OpenAPIRuntimeTests", dependencies: ["OpenAPIRuntime"], swiftSettings: swiftSettings),
     ]
 )
+
+// ---    STANDARD CROSS-REPO SETTINGS DO NOT EDIT   --- //
```
czechboy0, line 38: Are we missing a START: here?

Package.swift
czechboy0: Yeah, the other repos didn't get reformatted, so I wonder if we could undo it here: https://github.com/swift-server/swift-openapi-async-http-client/pull/43/files

rnro: Sorry, that was accidental - I've reverted the formatting changes

Package.swift
rnro: This is to try and have the lengths of the two blocks the same and have the text centered so it isn’t too ugly. We can revisit the exact structure of this the next time the common block is updated

PR #134, opened 2024-12-17
Adopt SE-409

czechboy0:
### Motivation

- https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md

### Modifications

Added explicitly access modifiers on imports.

### Result

Made it more explicit which types we're importing publicly and are used in our own API.

### Test Plan

Tests passed, will let CI ensure this works in all our supported Swift versions.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/Converter+Client.swift
+++ b/Sources/OpenAPIRuntime/Conversion/Converter+Client.swift
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 import Foundation
-import HTTPTypes
+public import HTTPTypes
 
 extension Converter {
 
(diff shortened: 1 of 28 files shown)


Sources/OpenAPIRuntime/Conversion/Converter+Client.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 import Foundation
```
FranzBusch, line 14: Isn't this API breaking? You were previously importing Foundation as `public` but with the changed flag it is now imported as `internal`. Since imports are leaky in Swift this would be API breaking.

PR #137, opened 2024-12-21
[CI] Enable Linux nightly main on PRs

czechboy0:
### Motivation

PR CI was missing testing with nighly Linux toolchains from main.

### Modifications

Enable nightly toolchain CI as well.

### Result

More test coverage on PRs.

### Test Plan

See this PR's CI if this works or if we need more changes.


czechboy0:
We have a bit of a cycle here: first we need to land this PR, without enable the integration test on nightly toolchains. Then we rekick https://github.com/apple/swift-openapi-generator/pull/709 and get that landed. And _then_ we can enable the integration test here again.

FranzBusch (approved):

PR #140, opened 2025-02-07
ClientError: don't call our to localizedDescription in description

weissi:
### Motivation

Fixes https://github.com/apple/swift-openapi-generator/issues/730 . LocalizedDescription doesn't print anything useful for Swift Errors. It just prints `ErrorType code 1` regardless what the underlying issue was.

That is unacceptable in production software.

### Modifications

Correct `description` to call `description`/`String(describing:)`.

### Result

Can debug software regardless of the transport.

### Test Plan

_[Describe the steps you took, or will take, to qualify the change - such as adjusting tests and manual testing.]_


czechboy0:
We're unlikely to take this PR, as this would expose error types that never opted into logging.

However, we will fix the forwarding logic to properly separate description and localizedDescription, tracked by https://github.com/apple/swift-openapi-generator/issues/730.

PR #143, opened 2025-02-24
Fix error description forwarding

czechboy0:
### Motivation

Fixes #730.

We were not correctly keeping `CustomStringConvertible` and `LocalizedError` conformances separate for wrapper errors like `ClientError` and `ServerError`. This lead to some user-thrown errors (in handlers, transports, and middlewares) to print less information than the error was actually providing (using a different method).

### Modifications

Properly untangle the two printing codepaths, and only call `localizedDescription` from the wrapper error's `errorDescription`.

Also made the `localizedDescription` strings a bit more user-friendly and less detailed, as in some apps these errors might get directly rendered by a UI component that calls `localizedDescription`.

### Result

Error logging should now match adopter expectations.

### Test Plan

Added unit tests for `{Client,Server}Error` printing methods.

### DIFF
--- a/Sources/OpenAPIRuntime/Conversion/ErrorExtensions.swift
+++ b/Sources/OpenAPIRuntime/Conversion/ErrorExtensions.swift
@@ -114,12 +114,18 @@ struct MultiError: Swift.Error, LocalizedError, CustomStringConvertible {
     var description: String {
         let combinedDescription =
             errors.map { error in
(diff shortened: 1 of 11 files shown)


czechboy0:
cc @weissi

simonjbeaumont (approved):
Looks good, thanks @czechboy0! 

Took me a minute to confirm that `prettyDescription` didn't get in the way again, but I believe this resolves the ask and we can avoid making use of localized description in pretty description going forward too.

PR #164, opened 2025-12-18
Add Wasm CI

czechboy0:
### Motivation

Allow generated code from OpenAPI to be used with Wasm.

### Modifications

Added CI to discover the current status.

### Result

Wasm CI, ensuring we don’t regress Wasm support once it works.

### Test Plan

Using this to see the current status.


diederich:
Worth bumping this to see where we are?

I have this [commit](https://github.com/apple/swift-openapi-runtime/commit/41fef2a17789fcfca8df7dd214bdfa063e799493) lying around which might help?

diederich:
Thanks for bumping! 🙏  Looks like the error I was seeing. Add the patch to this PR?

MaxDesiatov:
There are neither socket nor HTTP APIs available in WASI 0.1, besides the current OpenAPI infrastructure available on `main` isn't compatible with Embedded Swift. At least in the short term, compatibility with Embedded Swift and Web APIs available via JS interop in Swift for Wasm in the browser is a requirement, which this PR doesn't provide, so it's likely to be superseded by https://github.com/apple/swift-openapi-runtime/pull/187.

diederich:
Gotcha - makes sense. We did hack together a `fetch` based transport and got that to work. I *think* having build right now could allow that going forward?

PR #196, opened 2026-05-16
Deduplicate package Swift settings

fhwvtqdc2q-svg:
## Summary
- remove the initial per-target swiftSettings assignment from Package.swift
- rely on the existing package target loop to apply ExistentialAny, MemberImportVisibility, and InternalImportsByDefault once to every target

## Validation
- ran git diff --check
- verified Package.swift has a single ExistentialAny setting application path

Note: I could not run SwiftPM tests in this environment because Swift/Xcode is not installed on the Windows host.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,13 +14,6 @@
 //===----------------------------------------------------------------------===//
 import PackageDescription
 
-// General Swift-settings for all targets.
-let swiftSettings: [SwiftSetting] = [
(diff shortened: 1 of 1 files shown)


simonjbeaumont (approved):
Yep, this project adopted `ExistentialAny` before we were rolling it out in the automated compiler settings that we apply across a bunch of the repos. Thanks for the tidy up.

PR #197, opened 2026-05-16
Redact sensitive values in error descriptions

fhwvtqdc2q-svg:
## Summary
- redact sensitive HTTP header values when formatting runtime error descriptions
- redact common sensitive query parameters from request paths in formatted error descriptions
- add regression coverage for ClientError and ServerError descriptions

## Why
ClientError and ServerError include request/response metadata in their printable descriptions. Those descriptions can be logged by applications, and currently include raw header values such as Authorization, Cookie, and Set-Cookie, as well as query-string tokens. Redacting those values keeps the diagnostics useful while reducing accidental credential leakage.

## Testing
Not run locally because this environment is Windows and does not have Swift/Xcode installed. A macOS GitHub Actions validation run will be triggered on the fork branch.

### DIFF
--- a/Sources/OpenAPIRuntime/Interface/CurrencyTypes.swift
+++ b/Sources/OpenAPIRuntime/Interface/CurrencyTypes.swift
@@ -74,14 +74,78 @@ extension ServerRequestMetadata: CustomStringConvertible {
 }
 
 extension HTTPFields: PrettyStringConvertible {
+    private static var sensitiveHeaderNames: Set<String> {
+        [
+            "authorization",
+            "proxy-authorization",
+            "cookie",
+            "set-cookie",
+            "x-api-key",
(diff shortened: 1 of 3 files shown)


fhwvtqdc2q-svg:
Validation note: because this Windows environment cannot run Swift/Xcode locally, I validated the same code on a fork-only validation branch using macOS GitHub Actions. The validation branch adds only a workflow file on top of this PR's code. Run: https://github.com/fhwvtqdc2q-svg/swift-openapi-runtime/actions/runs/25954832217

FranzBusch:
I don't think this is the right fix. Swift `descriptions` are used for debugging as well. If you log a value you should only log the parts of the value that are safe to log. Logging an entire Swift value is always risky and can leak potential PII into logs.

PR #204, opened 2026-07-18
Adopt MultipartKit

ptoffy:
### Motivation

This PR adopts MultipartKit in favour of the custom multipart implementations that the package currently vends. It enables the runtime to remove a lot of code and instead rely on a well known and used package for its internals.
It's currently a draft because it relies on https://github.com/vapor/multipart-kit/pull/145/changes.
Let me know if this is a direction you'd like to explore!

### Modifications

Replaces the custom multipart code with MultipartKit's.

### Result

No more custom multipart code.

### Test Plan

The tests that tested the removed code were removed as well, no new tests were added as there's plenty of existing ones that test usage of the multipart code already


czechboy0:
Thanks for the PR, @ptoffy - I think this is definitely a direction we should explore. Unfortunately, we won't have the time to do all the necessary qualifications to take this PR for some time. Please leave the PR open, though, and once MultipartKit 5.0 ships, we can revisit later.

0xTim:
@czechboy0 if you get a chance to have a look at some point (no rush) that would be great. Just so we can make any breaking changes if there are any before the release if that unblocks the PR

PR #205, opened 2026-07-29
Fix semicolon parsing in `ContentDisposition`

kukushechkin:
### Motivation

Unconditional split at `;` splits `filename` value into multiple (usually broken) params.

### Modifications

Correctly track quotes in `ContentDisposition` param and inly split params at high-level semicolon.

### Result

Filenames with semicolons are parsed correctly.

### Test Plan

Unit tests.

### DIFF
--- a/Sources/OpenAPIRuntime/Base/ContentDisposition.swift
+++ b/Sources/OpenAPIRuntime/Base/ContentDisposition.swift
@@ -111,10 +111,12 @@ extension ContentDisposition: RawRepresentable {
         var isInsideQuotedString = false
         var isExpectingEscapedCharacter = false
         for character in rawValue {
-            switch character {
-            case _ where isExpectingEscapedCharacter:
+            if isExpectingEscapedCharacter {
                 current.append(character)
                 isExpectingEscapedCharacter = false
+                continue
+            }
+            switch character {
             case #"\"# where isInsideQuotedString:
                 current.append(character)
                 isExpectingEscapedCharacter = true
--- a/Sources/OpenAPIRuntime/Base/ContentDisposition.swift
+++ b/Sources/OpenAPIRuntime/Base/ContentDisposition.swift
@@ -110,10 +110,10 @@ extension ContentDisposition: RawRepresentable {
         var iterator = rawValue.makeIterator()
         while let character = iterator.next() {
             switch character {
-            case "\\" where isInsideQuotedString:
+            case #"\"# where isInsideQuotedString:
                 current.append(character)
                 if let escaped = iterator.next() { current.append(escaped) }
-            case "\"":
+            case #"""#:
                 isInsideQuotedString.toggle()
                 current.append(character)
             case ";" where !isInsideQuotedString:
@@ -128,9 +128,9 @@ extension ContentDisposition: RawRepresentable {
 
     /// Removes surrounding quotes and resolves backslash-escaped characters in a parameter value.
     private static func unquote(_ value: String) -> String {
-        guard value.count >= 2, value.first == "\"", value.last == "\"" else { return value }
-        return value.dropFirst().dropLast().replacingOccurrences(of: "\\\"", with: "\"")
-            .replacingOccurrences(of: "\\\\", with: "\\")
+        guard value.count >= 2, value.first == #"""#, value.last == #"""# else { return value }
+        return value.dropFirst().dropLast().replacingOccurrences(of: #"\""#, with: #"""#)
+            .replacingOccurrences(of: #"\\"#, with: #"\"#)
     }
 
     /// Wraps a parameter value in quotes, escaping backslashes and double quotes.
--- a/Sources/OpenAPIRuntime/Base/ContentDisposition.swift
+++ b/Sources/OpenAPIRuntime/Base/ContentDisposition.swift
@@ -103,25 +103,32 @@ struct ContentDisposition: Hashable {
 extension ContentDisposition: RawRepresentable {
 
     /// Splits a value into top-level components on `;`, without splitting inside a quoted value.
+    ///
+    /// Returns an empty array if the value has an unterminated quoted string, or a trailing
+    /// unescaped backslash inside a quoted string.
     private static func splitIntoTopLevelComponents(_ rawValue: String) -> [String] {
         var components: [String] = []
         var current = ""
         var isInsideQuotedString = false
-        var iterator = rawValue.makeIterator()
-        while let character = iterator.next() {
+        var isExpectingEscapedCharacter = false
+        for character in rawValue {
             switch character {
+            case _ where isExpectingEscapedCharacter:
+                current.append(character)
+                isExpectingEscapedCharacter = false
             case #"\"# where isInsideQuotedString:
                 current.append(character)
-                if let escaped = iterator.next() { current.append(escaped) }
+                isExpectingEscapedCharacter = true
             case #"""#:
-                isInsideQuotedString.toggle()
                 current.append(character)
+                isInsideQuotedString.toggle()
             case ";" where !isInsideQuotedString:
                 if !current.isEmpty { components.append(current) }
                 current = ""
             default: current.append(character)
             }
         }
+        guard !isInsideQuotedString, !isExpectingEscapedCharacter else { return [] }
         if !current.isEmpty { components.append(current) }
         return components.map { $0.trimmingLeadingAndTrailingSpaces }
     }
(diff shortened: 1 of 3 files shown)


Sources/OpenAPIRuntime/Base/ContentDisposition.swift
```diff
@@ -102,21 +102,57 @@ struct ContentDisposition: Hashable {
...
+    /// Removes surrounding quotes and resolves backslash-escaped characters in a parameter value.
+    private static func unquote(_ value: String) -> String {
+        guard value.count >= 2, value.first == "\"", value.last == "\"" else { return value }
+        return value.dropFirst().dropLast().replacingOccurrences(of: "\\\"", with: "\"")
+            .replacingOccurrences(of: "\\\\", with: "\\")
+    }
+
+    /// Wraps a parameter value in quotes, escaping backslashes and double quotes.
+    private static func quote(_ value: String) -> String {
+        "\"" + value.replacingOccurrences(of: "\\", with: "\\\\").replacingOccurrences(of: "\"", with: "\\\"") + "\""
```
czechboy0, line 138:
```suggestion
        #"""# + value.replacingOccurrences(of: #"\"#, with: #"\\"#).replacingOccurrences(of: #"""#, with: #"\""#) + #"""#
```

Sources/OpenAPIRuntime/Base/ContentDisposition.swift
```diff
@@ -102,21 +102,57 @@ struct ContentDisposition: Hashable {
...
 
+    /// Splits a value into top-level components on `;`, without splitting inside a quoted value.
+    private static func splitIntoTopLevelComponents(_ rawValue: String) -> [String] {
+        var components: [String] = []
+        var current = ""
+        var isInsideQuotedString = false
+        var iterator = rawValue.makeIterator()
+        while let character = iterator.next() {
+            switch character {
+            case "\\" where isInsideQuotedString:
```
czechboy0, line 113:
```suggestion
            case #"\"# where isInsideQuotedString:
```

Sources/OpenAPIRuntime/Base/ContentDisposition.swift
```diff
@@ -102,21 +102,57 @@ struct ContentDisposition: Hashable {
...
+                current.append(character)
+                if let escaped = iterator.next() { current.append(escaped) }
+            case "\"":
```
czechboy0, line 116:
```suggestion
            case #"""#:
```

Sources/OpenAPIRuntime/Base/ContentDisposition.swift
```diff
@@ -102,21 +102,57 @@ struct ContentDisposition: Hashable {
...
+            default: current.append(character)
+            }
+        }
+        if !current.isEmpty { components.append(current) }
+        return components.map { $0.trimmingLeadingAndTrailingSpaces }
+    }
+
+    /// Removes surrounding quotes and resolves backslash-escaped characters in a parameter value.
+    private static func unquote(_ value: String) -> String {
+        guard value.count >= 2, value.first == "\"", value.last == "\"" else { return value }
```
czechboy0, line 131:
```suggestion
        guard value.count >= 2, value.first == #"""#, value.last == #"""# else { return value }
```

Sources/OpenAPIRuntime/Base/ContentDisposition.swift
```diff
@@ -102,21 +102,57 @@ struct ContentDisposition: Hashable {
...
+        return value.dropFirst().dropLast().replacingOccurrences(of: "\\\"", with: "\"")
+            .replacingOccurrences(of: "\\\\", with: "\\")
```
czechboy0, line 133:
```suggestion
        return value.dropFirst().dropLast().replacingOccurrences(of: #"\""#, with: #"""#)
            .replacingOccurrences(of: #"\\"#, with: #"\"#)
```

Sources/OpenAPIRuntime/Base/ContentDisposition.swift
```diff
@@ -102,21 +102,57 @@ struct ContentDisposition: Hashable {
...
+    private static func splitIntoTopLevelComponents(_ rawValue: String) -> [String] {
+        var components: [String] = []
+        var current = ""
+        var isInsideQuotedString = false
+        var iterator = rawValue.makeIterator()
+        while let character = iterator.next() {
+            switch character {
+            case "\\" where isInsideQuotedString:
+                current.append(character)
+                if let escaped = iterator.next() { current.append(escaped) }
```
czechboy0, line 115: Can you help me understand this section better? I don't quite follow - we don't transform the characters, we just add them back as they were.

Tests/OpenAPIRuntimeTests/Multipart/Test_MultipartValidationSequence.swift
```diff
@@ -271,6 +271,30 @@ final class Test_MultipartValidationSequenceStateMachine: Test_Runtime {
         XCTAssertEqual(stateMachine.next(parts[0]), .emitError(.receivedMultipleValuesForSingleValuePart("name")))
     }
 
+    func testFilenameContainingQuoteAndSemicolon() throws {
```
czechboy0, line 274: What does this test on top of what you added to `Test_ContentDisposition.swift`?

Sources/OpenAPIRuntime/Base/ContentDisposition.swift
kukushechkin: Yes, here we add the next symbol as is, whatever is escaped by that "\\" **inside** the quoted string, so we correctly handle "\"", it shouldn't end the quoted string.

Tests/OpenAPIRuntimeTests/Multipart/Test_MultipartValidationSequence.swift
kukushechkin: One is testing `ContentDisposition`, another one `MultipartValidationSequence`.

czechboy0: I suggest we just remove this test, the other one is enough. This one is too indirect.

kukushechkin: Removed.

Sources/OpenAPIRuntime/Base/ContentDisposition.swift
```diff
@@ -102,21 +102,57 @@ struct ContentDisposition: Hashable {
...
+    private static func splitIntoTopLevelComponents(_ rawValue: String) -> [String] {
+        var components: [String] = []
+        var current = ""
+        var isInsideQuotedString = false
+        var iterator = rawValue.makeIterator()
+        while let character = iterator.next() {
+            switch character {
+            case #"\"# where isInsideQuotedString:
+                current.append(character)
+                if let escaped = iterator.next() { current.append(escaped) }
```
czechboy0:
```suggestion
                guard let escaped = iterator.next() else { return [] }
                current.append(escaped)
```

This has bothered me, that we're silently parsing strings with unbalanced quotes. This at least refuses to parse it altogether, which I think is better.

czechboy0 (commented):
Just one more nit, and please rerun format, the CI is failing.

Sources/OpenAPIRuntime/Base/ContentDisposition.swift
kukushechkin: This is a good catch. But this will only reject if quotes are unbalanced, but the terminating char is `\`. We should instead check if at the end of the string process we're still inside the quotes. I've updated the state machine (that inner `.next()` has been itching my eye anyway) and added a check at the end.

Sources/OpenAPIRuntime/Base/ContentDisposition.swift
```diff
@@ -102,21 +102,64 @@ struct ContentDisposition: Hashable {
...
+    private static func splitIntoTopLevelComponents(_ rawValue: String) -> [String] {
+        var components: [String] = []
+        var current = ""
+        var isInsideQuotedString = false
+        var isExpectingEscapedCharacter = false
+        for character in rawValue {
+            switch character {
+            case _ where isExpectingEscapedCharacter:
+                current.append(character)
+                isExpectingEscapedCharacter = false
```
czechboy0:
```suggestion
            if isExpectingEscapedCharacter {
                current.append(character)
                isExpectingEscapedCharacter = false
                continue
            }
            switch character {
```

Equivalent, hiding that if into the switch confused me a bit :)

czechboy0 (changes requested):
One more nit, sorry

czechboy0 (approved):

PR #206, opened 2026-07-29
Strip newlines when parsing events

kukushechkin:
### Motivation

The `id` and `event` fields aren't normalized the same way `data` already is, so they could end up split across multiple lines.

### Modifications

Strip newlines from `id` and `event`.

### Result

`id` and `event` each is serialized into a single line.

### Test Plan

Unit test.

### DIFF
--- a/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsEncoding.swift
+++ b/Sources/OpenAPIRuntime/EventStreams/ServerSentEventsEncoding.swift
@@ -167,8 +167,15 @@ extension ServerSentEventsSerializationSequence.Iterator {
                         buffer.append(contentsOf: value.utf8)
                         buffer.append(ASCII.lf)
                     }
-                    if let id = value.id { encodeField(name: "id", value: id) }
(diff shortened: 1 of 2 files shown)


Sources/OpenAPIRuntime/EventStreams/ServerSentEventsEncoding.swift
```diff
@@ -167,8 +167,14 @@ extension ServerSentEventsSerializationSequence.Iterator {
                         buffer.append(contentsOf: value.utf8)
                         buffer.append(ASCII.lf)
                     }
-                    if let id = value.id { encodeField(name: "id", value: id) }
-                    if let event = value.event { encodeField(name: "event", value: event) }
+                    // Field values must not contain any newline characters, as they're used to
```
czechboy0:
```suggestion
                    // `id` and `event` values must not contain any newline characters, as they're used to
```

czechboy0 (commented):
Just a nit, lgtm otherwise - thanks @kukushechkin 🙏

czechboy0 (approved):

PR #211, opened 2026-09-18
Add OTel Logging Middleware

bauer-andreas:
### Motivation

Instrumentation support varies based on the transport. Especially current Client Transport lack any sort of middleware. Therefore, it is useful to leverage the OpenAPI middleware system to provide constant emission of telemetry across client and server transports. The Open Telemetry Semantic Conventions is a widely established convention for metadata emission.

### Modifications

* Add a new `ClientOTelLoggingMiddleware` and `ServerOTelLoggingMiddleware` following Open Telemetry Semantic Conventions for HTTP
* Introduce a new default-enabled `Logging` and `OTelSemanticConvention` trait that guards the two new dependencies on swift-log and swift-otel-semantic-conventions.

### Result

Consistent request and response logging across server and client implementations following OTel Semantic Conventions.

### Test Plan

Added two new unit test suites that exercise both of the middleware using an in-memory logger that is injected via the new TaskLocal Logger form swift-log.


czechboy0:
This is great, @bauer-andreas! I've wanted us to have telemetry middlewares built-in, so thanks for getting this started.

Since this has a pretty large API surface, and there are various smaller decisions that would benefit from more discussion, would you be open to writing up a [proposal](https://swiftpackageindex.com/apple/swift-openapi-generator/1.13.1/documentation/swift-openapi-generator/proposals)? We don't normally accept non-trivial new API without one. 🙏

czechboy0 (changes requested):
Marking as needs-work for now, as we'll need to run a proposal for this first.

czechboy0:
Also cc @simonjbeaumont and @slashmo for visibility and additional thoughts.

